PR #2, opened 2020-07-16
Migrate dynatrace-oneagent-operator chart v0.8.0

PennyScissors:
**Patch includes:**
Add reference to local copy of logo.png
Add partner annotation to Chart.yaml
Move questions.yaml, app-readme.md, and logo.png to overlay folder

The files `questions.yaml`, `app-readme.md`, and `logo.png` were moved to the overlay directory because our new way of packaging charts uses the partner's upstream chart repository rather than the version of their chart in our repositories (`rancher/chart` and `rancher/helm3-charts`). Their upstream shouldn't include these as they are Rancher specific and should be kept separate in the overlay directory. We usually ask partners to create those before submitting a PR to our repositories, but this partner happened to include them in their upstream chart too (not a common occurrence).


StrongMonkey:
@PennyScissors The patch file should not show patches for overlay files. I believe @aiyengar2 did some work on the script and maybe this is not synced with the latest dev-charts. Can you check with him?

aiyengar2:
> @PennyScissors The patch file should not show patches for overlay files. I believe @aiyengar2 did some work on the script and maybe this is not synced with the latest dev-charts. Can you check with him?

@StrongMonkey this shouldn't be related to the script I pushed. This PR is manually deleting the `questions.yml` file in the upstream chart and replacing it with an `overlay/questions.yml` version of the file instead

PennyScissors:
@StrongMonkey @aiyengar2 The patch shows patched overlay files because they were replaced in the chart for the ones in overlay. I updated the description and commit message with more information as for why they were replaced.

packages/dynatrace-oneagent-operator/overlay/questions.yml
```diff
@@ -0,0 +1,179 @@
+categories:
+- APM
+- Monitoring
+labels:
```
deniseschannon, line 4: remove this label as we will move it into Chart.yaml as an annotation

StrongMonkey:
@PennyScissors Can you copy `scripts` and `.github` folder from rancher/dev-charts and sync with parnter-charts? We made a bunch changes in CI and would like partner-chart to adopt this. After that Can you regenerate your patch files?

deniseschannon:
@PennyScissors let's merge this!

PR #5, opened 2020-08-02
Sync from rancher/charts

StrongMonkey:


packages/dynatrace-oneagent-operator/package.yaml
```diff
@@ -1,5 +1,5 @@
 url: https://github.com/Dynatrace/helm-charts
 subdirectory: dynatrace-oneagent-operator
-commit: dynatrace-oneagent-operator-v0.8.0
```
StrongMonkey, line 3: Change this to commit since we don't fetch tag to improve the speed

PR #13, opened 2020-08-04
Migrate traefik v8.9.1

PennyScissors:
**NEEDS FIX:**
- Can't access dashboard when enabled in questions by setting `ingressRoute.dashboard.enabled: true` (default is enabled).

Update app-readme based on their latest README.md
Updates in question.yaml to align with new chart values:
- Remove partner and role labels
- Remove categories
- Refactor variable names
- Update default image tag
- Remove debug mode as it's no longer available. Logging level can be
configured by passing flags to `additionalArguments`.
- Remove acme and ssl values as they are no longer available.

### DIFF
--- a/packages/traefik/overlay/questions.yml
+++ b/packages/traefik/overlay/questions.yml
@@ -1,0 +1,121 @@
+categories:
+- Proxy
+- Loadbalancer
+labels:
+  io.cattle.role: project
+  io.rancher.certified: partner
+questions:
+- variable: defaultImage
+  default: true
+  description: "Use default Docker image"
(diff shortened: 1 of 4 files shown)


packages/traefik/overlay/questions.yml
```diff
@@ -0,0 +1,48 @@
+categories:
+- Proxy
+- Loadbalancer
+labels:
+  io.cattle.role: project
+  io.rancher.certified: partner
```
deniseschannon, line 6: we should remove the partner label from the question.yml

PR #22, opened 2020-08-07
Migrate kong v1.8.0

PennyScissors:
**NEEDS FIX**
Installing CRDs by enabling the option in questions will result in the following error:
```
Failed to install app kong. Error: rendered manifests contain a resource that already exists. Unable to continue with install: CustomResourceDefinition "kongclusterplugins.configuration.konghq.com" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "kong"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "kong"
```

 Update questions
- Remove role and partner labels
- Remove categories


PennyScissors:
Closed because they fixed the issue in a new version of the chart.

PR #34, opened 2020-09-08
Add create partner chart script and update README

PennyScissors:
This pull request includes:
- Sync scripts with rancher/charts dev-v2.5

- Add script that creates a chart directory to simplify the chart submission process for new partners
User provides `URL` (required) and `CHART` (optional, it tries creating one from URL first) and the following tree structure is generated:
```text
partner-charts
└── packages
    └── ${CHART}
        ├── ${CHART}.patch 
        ├── overlay
        │   ├── app-readme.md  # empty, required for partner charts
        │   └── questions.yaml  # empty, required for partner charts
        └── package.yaml  # url is populated and packageVersion set to 00 as default
```
- Update README to be more partner charts specific


scripts/create-partner-chart
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
```
cbron, line 1: So in general I'm very hesitant to add anything to the scripts stuff, but I suppose since this is actually just a partner charts folder layout generator thats used one time it should be ok.

scripts/create-partner-chart
```diff
@@ -0,0 +1,59 @@
...
+# Verify required arguments
+if [[ ! $URL ]]; then
+    echo "Error: Missing required argument URL"
+    echo $usage
+    exit 1
+fi
+
+# Verify URL is valid and references a tgz resource
+if [[ ! $URL =~ https?://.*\.tgz ]]; then
+    echo "Error: Invalid URL form"
```
cbron, line 15: Should specify that its to a tgz, they might put index.yaml

scripts/create-partner-chart
```diff
@@ -0,0 +1,59 @@
...
+    exit 1
+fi
+curl -fsI $URL &>/dev/null && ok=1
+if [[ ! $ok ]]; then
+    echo "Error: No resource found at $URL"
+    exit 1
+fi
+
+# Try trimming tgz basename to generate a name for chart dir if one wasn't provided
```
cbron, line 24: I'd just force them to supply a name because the url could get weird

scripts/create-partner-chart
```diff
@@ -0,0 +1,59 @@
...
+fi
+
+# Verify a chart with the same name doesn't exist already
+chart_dir=packages/$CHART
+if [ -d $chart_dir ]; then
+    echo "Error: $chart_dir already exists"
+    exit 1
+fi
+
+# Create directory and patch for new chart
```
cbron, line 44: Doesn't need a patch file, that gets generated if it doesn't exist. Unless this is a partner specific thing.

README.md
```diff
@@ -1,126 +1,188 @@
 # partner-charts
+This repository is reserved for partner charts in the Rancher's v2.5+ catalog.
 
-This is an experimental project to automate system charts and adding overlays files on top of it.
+As part of this catalog, all charts will benefit of a cloud native packaging system
```
cbron, line 4: Are we forcing helm3 on partner charts now ?

README.md
```diff
@@ -1,126 +1,188 @@
...
+Rancher specific modifications and adding overlay files on top of it.
 
-Charts will have two categories: 
+## Requirements
+* Chart must be Helm v3 compatible
 
-1. Charts that Rancher created and maintained (**Rancher original**).
+    A common cause for Helm v3 incompatibility is using the `crd-install` hook provided in Helm 2 for 
+    the installation of CRDs. This hook was removed in v3 in favor of a `crds` directory where your 
+    crds should now reside. Templating and upgrading crds is also no longer supported.
```
cbron, line 13: Upgrading is supported.

cbron: Wait, are partner charts running through our chart split method or not ?

README.md
```diff
@@ -1,126 +1,188 @@
...
+```
+make charts CHART={CHART_NAME}
+```
 
-Generate tarball for each charts. Use `CHART=${NAME}` for specific chart.
+#### Test modified chart
 
-`make patch`: 
+An easy way to test your chart is by hosting the `assets` directory (Make sure you generate the tarball first) using 
+`python -m SimpleHTTPServer`, expose it using `ngrok`, and use the `ngrok` URL to add the chart as a custom catalog in Rancher.
```
cbron, line 170: Potentially easier way is to just `helm install assets/x.tgz` with the kubeconfig set to that cluster. Not sure which route we want them taking, I guess they don't get questions.yaml that way.

README.md
```diff
@@ -1,126 +1,188 @@
...
+```
 
-`make mirror`: 
+At this point, you are ready to submit a pull request.
+All the contents of you chart directory should be commited, except the `charts` directory.
 
-Run image mirroring scripts.(Experimental)
+#### Hotfix
+Our CI doesn't allow modifying existing charts, therefore bumping a chart's version after modifying it is required.
+We understand there are cases a hotfix might be better suited, and although is not recommended, the option to override
```
cbron, line 186: spacing seems weird here

scripts/create-partner-chart
PennyScissors: I'll update it to include that.

scripts/create-partner-chart
PennyScissors: I know it gets created, but I figured is easier if they don't even have to worry about whether they need to create it or not. Since all partner charts will require a patch to add the partner annotation, I figured it would be good to just create it for them at the very first step.

README.md
PennyScissors: Yes, because 2.5+ uses Helm 3 only (as far as I know) and right now if they submit a Helm 2 chart we would have to worry about whether is Helm 3 compatible before migrating it to the new catalog.

README.md
PennyScissors: Last I heard, partners will not run through our chart split method (Might need @deniseschannon to confirm). How they handle their CRDs is their responsibility.

README.md
PennyScissors: Is important they can test their chart with the questions.yaml and app-readme.md included because that's usually where new partners need help the most.

README.md
PennyScissors: I'll fix it.

scripts/create-partner-chart
PennyScissors: I check the URL first for these:
1. URL matches the form http://www.name.com/chart-name-v1.0.0.tgz
2. References a resource (checks with curl)
3. Doesn't contain any dots (in case it has extra versions like chart-name-v1.0.0-beta.tgz, etc)

If that fails, a message shows up and the user is forced to supply the name, but I don't expect that to happen as much because we are asking them to use helm repos which usually include GA releases only and standard semantic versioning.

The reason I don't just force them to supply one and rather try creating it first is because they might not look at other charts and just supply a random name with a different case(pascal, camel, underscore, etc), so I at least try to create it for them first. At the end it shouldn't really matter since either case would easily get caught in the PR review.

If you don't agree, I have no problems updating it to force them to supply it. I'm just going based on my experience with partners so far.

scripts/create-partner-chart
PennyScissors: If we are coming up with something new soon, I would definitely prefer waiting on that before we ask partners to add charts here by themselves and just migrate them for them. We are asking them to use the rancher/helm3-charts repo now, which means their charts need to be Helm 3 compatibl and that should make migrating them as easy as populating the package.yaml, add partner annotation, and copy overlays.

cbron: I think package.yaml and annotations should stay roughly the same in future, we are just rebuilding framework that builds them out. But if we make a decision around adding or removing annotation or change to package.yaml format, we are going to have to have that done on all partner charts.

README.md
cbron: Ok then it would not be

cbron (commented):
I don't see new commit ?

PennyScissors:
@cbron I'll let you know once I update this. I'm currently working on a task of higher priority so this is on hold.

PR #36, opened 2020-09-10
[dev-v2.5] move to main as primary branch

cbron:
Pushing into 2.5 branch is necessary so that build doesn't overwrite the main one

Rancher PR merge: https://github.com/rancher/rancher/pull/28778


cbron:
We can also just delete this branch and then this PR is unnecessary

PR #47, opened 2020-10-15
Update README to be partner specific

PennyScissors:

PR #48, opened 2020-10-16
Update readme and sync scripts with rancher/charts

PennyScissors:
Readme was already reviewed in PR https://github.com/rancher/partner-charts/pull/47, but we will no longer use branch `dev-v2.5` for partners so `main-source` needs to be updated.


cbron (approved):
This will likely regenerate all the charts

PR #49, opened 2020-10-21
Update README to include maintenance steps

PennyScissors:


README.md
```diff
@@ -170,7 +170,35 @@ questions:
...
+will pull in the new upstream chart and apply the patch if one exists.
+
+For example, an existing package tracking an upstream chart `url: https://example.com/helm-repo/chart-v0.1.2.tgz`
+can be updated to track the new `url: https://example.com/helm-repo/chart-v0.1.3.tgz`, and a new package 
+`chart-v0.1.300.tgz` will be generated.
+
+```yaml
+url: https://example.com/helm-repo/chart-v0.1.3.tgz
+packageVersion: 00
+```
```
aiyengar2, line 190: There should be a warning here that they might need to run `make patch` manually to ensure that the patch can be applied on the new upstream; if not, there's currently no method for rebasing to a new upstream when the patch gets broken as a result (e.g. rancher-monitoring can't be updated easily for this reason).

Most likely, unless this is a chart like `fleet` that doesn't require patches, the patch will be broken for a new upstream.

aiyengar2 (changes requested):

README.md
```diff
@@ -170,7 +170,35 @@ questions:
...
 
-### 6. Test your changes
+### 6. Update package to track new upstream (Maintenance)
+
+There are two ways you can update a package, one is to track a new updated upstream chart
+and the other is to do small modifications/fixes.
+
+#### Update package to track a new upstream chart
+
+Update the `url` in `package.yaml` to reference the new upstream chart and `make prepare CHART={CHART_NAME}`
```
aiyengar2, line 180: We should add a note about resetting the `packageVersion` as well when you rebase to a new upstream.

aiyengar2 (changes requested):
minor comment, should be good to merge after.

aiyengar2 (approved):

PR #66, opened 2021-01-19
Add Nutanix CSI Storage

tuxtof:


PennyScissors:
Hello @tuxtof, have you tested this chart in Rancher 2.5? I tried installing it, but it gets stuck deploying a daemonset. Is there any set up that needs to be done before I install the chart?
```
DaemonSet is not ready: default/csi-node-ntnx-plugin. 0 out of 1 expected pods are ready
```

PR #70, opened 2021-01-29
HPE CSI Driver for Kubernetes Helm v3 Chart 1.4.2

datamattsson:
Signed-off-by: Michael Mattsson <michael.mattsson@hpe.com>

### DIFF
--- a/packages/hpe-csi-driver/hpe-csi-driver.patch
+++ b/packages/hpe-csi-driver/hpe-csi-driver.patch
@@ -1,17 +1,29 @@
 diff -x '*.tgz' -x '*.lock' -uNr packages/hpe-csi-driver/charts-original/Chart.yaml packages/hpe-csi-driver/charts/Chart.yaml
 --- packages/hpe-csi-driver/charts-original/Chart.yaml
 +++ packages/hpe-csi-driver/charts/Chart.yaml
-@@ -17,3 +17,6 @@
-   - CentOS
-   - Ubuntu
-   - RHEL
+@@ -7,9 +7,6 @@
+ - HPE
+ - Storage
+ - StorageClass
+-- CentOS
+-- Ubuntu
+-- RHEL
+ maintainers:
+ - email: hpe-containers-dev@hpe.com
+   name: raunakkumar
+@@ -17,3 +14,6 @@
+ sources:
+ - https://scod.hpedev.io/csi_driver
+ version: 1.4.2
 +annotations:
 +  catalog.cattle.io/certified: partner
 +  catalog.cattle.io/release-name: hpe-csi-driver
 diff -x '*.tgz' -x '*.lock' -uNr packages/hpe-csi-driver/charts-original/questions.yml packages/hpe-csi-driver/charts/questions.yml
 --- packages/hpe-csi-driver/charts-original/questions.yml
 +++ packages/hpe-csi-driver/charts/questions.yml
-@@ -1,5 +1,3 @@
+@@ -1,7 +1,3 @@
+-labels:
+-  io.rancher.certified: partner
 -categories:
 -- storage
  questions:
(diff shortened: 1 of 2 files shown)


PennyScissors:
Hello @datamattsson, assuming this chart is the same as the other in rancher/helm3-charts, and taking into consideration that you mentioned that beyond v1.4.0 you will only use this repository for your releases, there are a few things worth pointing out.

1. This catalog is available in the new Cluster Explorer UI only, which lets you modify your chart's `values.yaml` in the UI directly. For that reason, the `questions.yaml` is optional. If you want users to have the nicer configuration in the UI with the set amount of questions you provided, keep the `questions.yaml`. Otherwise, you can remove it and they will be presented with the option to see the `values.yaml` content and configure it as they like before installing the chart.

2. Your `questions.yaml` and `app-readme.md` files are packaged within your chart in your hosted helm repo. This setup works okay, but is not ideal because they are Rancher specific files and is not really the convention. As you continue maintenance of your chart and need to update those files, it will become a bit confusing to do so. You would need to either patch the changes in here, which is not desirable as it will clutter the patch and it's much harder to review the longer it gets, or make the changes in your own repository and update the reference here. Ideally, the patch for a partner chart will only contain the required annotations. The best approach to fix this would be to delete those files from upstream so the tgz you reference in `package.yaml` doesn't contain them, and copy them into an `overlay/` directory as shown in the example below. If you decided to have users modify your `values.yaml` instead (refer to point 1), then don't include your `questions.yaml` in `overlay/` . Files in `overlay/` are copied into the chart we generate before publishing, so modification can be done to the files directly without a need for saving them in the patch.
```
|____packages/
|    |____hpe-csi-driver/
|    |    |____overlay/
|    |    |    |____questions.yaml
|    |    |    |____app-readme.md
|    |    |____package.yaml
|    |    |____hpe-csi-driver.patch
```

3. Optionally, you can deprecate the chart in helm3-chart by setting up a constraint so that it's hidden from the catalog (prevents new installations only, not using it) in say Rancher v2.5.5+ and that way you can effectively deprecate it without affecting current users and shift your focus to maintaining this chart only. If this is something you would like to do, please let me know and I'll comment in the other PR what you need to do exactly.

PennyScissors (commented):
Hello @datamattsson, assuming this chart is the same as the other in rancher/helm3-charts, and taking into consideration that you mentioned that beyond v1.4.0 you will only use this repository for your releases, there are a few things worth pointing out.

1. This catalog is available in the new Cluster Explorer UI only, which lets you modify your chart's `values.yaml` in the UI directly. For that reason, the `questions.yaml` is optional. If you want users to have the nicer configuration in the UI with the set amount of questions you provided, keep the `questions.yaml`. Otherwise, you can remove it and they will be presented with the option to see the `values.yaml` content and configure it as they like before installing the chart.

2. Your `questions.yaml` and `app-readme.md` files are packaged within your chart in your hosted helm repo. This setup works okay, but is not ideal because they are Rancher specific files and is not really the convention. As you continue maintenance of your chart and need to update those files, it will become a bit confusing to do so. You would need to either patch the changes in here, which is not desirable as it will clutter the patch and it's much harder to review the longer it gets, or make the changes in your own repository and update the reference here. Ideally, the patch for a partner chart will only contain the required annotations. The best approach to fix this would be to delete those files from upstream so the tgz you reference in `package.yaml` doesn't contain them, and copy them into an `overlay/` directory as shown in the example below. If you decided to have users modify your `values.yaml` instead (refer to point 1), then don't include your `questions.yaml` in `overlay/` . Files in `overlay/` are copied into the chart we generate before publishing, so modification can be done to the files directly without a need for saving them in the patch.
```
|____packages/
|    |____hpe-csi-driver/
|    |    |____overlay/
|    |    |    |____questions.yaml
|    |    |    |____app-readme.md
|    |    |____package.yaml
|    |    |____hpe-csi-driver.patch
```

3. Optionally, you can deprecate the chart in helm3-chart by setting up a constraint so that it's hidden from the catalog (prevents new installations only, not using it) in say Rancher v2.5.5+ and that way you can effectively deprecate it without affecting current users and shift your focus to maintaining this chart only. If this is something you would like to do, please let me know and I'll comment in the other PR what you need to do exactly.

datamattsson:
For the 1.4.x of the chart we want to keep the app readme and questions file in upstream repo. For 1.5.x, those files will be removed and we will not provide further updates to helm3-charts repo.

PennyScissors (approved):

PR #71, opened 2021-01-29
Partner chart for k8s-triliovault-operator v2.0.2

bhagirathhapse:
The PR consists of:
1. package.yaml file mentioning helm package tgz file
2. overlay directory containing app-readme.md and questions.yaml
3. k8s-triliovault-operator.patch file after adding partner annotations in Charts.yaml file


packages/k8s-triliovault-operator/package.yaml
```diff
@@ -0,0 +1,2 @@
+url: https://raw.githubusercontent.com/trilioData/triliovault--operator/master/k8s-triliovault-operator-v2.0.2.tgz
+packageVersion:
```
PennyScissors, line 2: This should be: `packageVersion: 00`

packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
```diff
@@ -0,0 +1,32 @@
+diff -x '*.tgz' -x '*.lock' -uNr packages/k8s-triliovault-operator/charts-original/Chart.yaml packages/k8s-triliovault-operator/charts/Chart.yaml
+--- packages/k8s-triliovault-operator/charts-original/Chart.yaml
++++ packages/k8s-triliovault-operator/charts/Chart.yaml
+@@ -1,6 +1,7 @@
+ apiVersion: v1
+ appVersion: v2.0.2
+-description: K8s-TrilioVault-Operator is an operator designed to manage the K8s-TrilioVault Application Lifecycle.
++description: K8s-TrilioVault-Operator is an operator designed to manage the K8s-TrilioVault
++  Application Lifecycle.
```
PennyScissors, line 9: Not sure if this was intentional. If it was accidental, please remove it to avoid cluttering the patch, as it should only contain Rancher specific changes like the annotations below.

packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
```diff
@@ -0,0 +1,32 @@
...
++  catalog.cattle.io/certified: partner
++  catalog.cattle.io/release-name: k8s-triliovault-operator
+diff -x '*.tgz' -x '*.lock' -uNr packages/k8s-triliovault-operator/charts-original/app-readme.md packages/k8s-triliovault-operator/charts/app-readme.md
+--- packages/k8s-triliovault-operator/charts-original/app-readme.md
++++ packages/k8s-triliovault-operator/charts/app-readme.md
+@@ -0,0 +1,4 @@
++# TrilioVault for Kubernetes
++
++[K8s-TrilioVault-Operator](https://trilio.io) is an operator designed to manage the K8s-TrilioVault Application 
++Lifecycle.
```
PennyScissors, line 27: These changes shouldn't be in the patch. Files in the overlay directory are added when generating the asset, so there is no need to include them in patch. Is possible you saved this in the patch accidentally by generating the assets and doing make patch again without a make clean to cleanup the changes generating the assets leave in your working directory.

packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
```diff
@@ -0,0 +1,32 @@
...
+diff -x '*.tgz' -x '*.lock' -uNr packages/k8s-triliovault-operator/charts-original/questions.yml packages/k8s-triliovault-operator/charts/questions.yml
+--- packages/k8s-triliovault-operator/charts-original/questions.yml
++++ packages/k8s-triliovault-operator/charts/questions.yml
+@@ -0,0 +1 @@
++questions:
```
PennyScissors, line 32: Same as `app-readme.md` (see other comment), shouldn't be in the patch.

Also, questions are used in the UI to provide the user with a nicer way to configure the chart. Please include at least a few questions to cover common use cases for your chart.

packages/k8s-triliovault-operator/package.yaml
bhagirathhapse: Okay, I will make this change

packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
bhagirathhapse: This was accidental. @PennyScissors Do we need .patch file even when we are uploading the partner chart for the very first time? I thought, we need patch file only when we update the helm chart to new release or new product version.
Correct me my above understanding is wrong, if not, can we remove the .patch file from this PR?

packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
bhagirathhapse: If we remove the patch file then this won't be a problem, if not then I will remove these unnecessary changes from patch file.

packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
bhagirathhapse: We don't have any questios for users to answer for helm chart so if required user can update the `values.yaml` and I will remove the `questions.yaml` from the PR

packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
PennyScissors: The patch itself is not required, but in the case of a partner chart you need it so that you can add the required annotations. The idea is the patch should only contain all Rancher specific changes that you need, so most partner charts only have the annotations in their patch.

packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
PennyScissors: The patch can't be removed for partner charts since you need the annotations.

packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
PennyScissors: Questions populate the chart's values.yaml. It is just a way to configure it through the UI rather than modify the file itself. If you would rather have users modify the chart's values.yaml, you can delete the questions from the overlay directory, and make sure the patch doesn't contain any traces of it.

PennyScissors (commented):
Still needs some minor changes

packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
bhagirathhapse: Thank you @PennyScissors . I have updated the files accordingly. Please check and let me know.

packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
bhagirathhapse: Got it. Added the changes to this PR.

packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
bhagirathhapse: I have removed the `questions.yaml` since user does not need to asnwer any questions and parameters present in the `values.yaml` and good for installation.

bhagirathhapse:
@PennyScissors I have implemented the minor changes you have suggested. Please review and let me know if we are good to move ahead.

PennyScissors (approved):

PR #72, opened 2021-01-29
Upgraded artifactory charts to version 4.7.6 (Artifactory 7.12.6)

vinayagg:


packages/artifactory-ha/artifactory-ha.patch
```diff
@@ -1,10 +1,32 @@
...
++- jfrog
++- devops
++maintainers:
++- email: installers@jfrog.com
++  name: Chart Maintainers at JFrog
++name: artifactory-ha
++sources:
++- https://bintray.com/jfrog/product/JFrog-Artifactory-Pro/view
++- https://github.com/jfrog/charts
++version: 4.7.6
```
PennyScissors, line 32: This shouldn't be in the patch. Is likely that when you changed the url in `package.yaml` to your new upstream, the patch didn't apply cleanly and these were created. Please remove it if you patch is working properly. This applies to https://github.com/rancher/partner-charts/pull/73 as well

PennyScissors (changes requested):

packages/artifactory-ha/artifactory-ha.patch
vinayagg: @PennyScissors Thanks for your review. This issue has been fixed in this PR and also in #73 . Please check.

PennyScissors (approved):

PR #74, opened 2021-02-05
Ignore charts directory for Github Pages build

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/31021

#### Why is this needed?

Currently, we serve URLs like http://partner-charts.rancher.io/charts/artifactory-ha/ but there's no need for us to do that. This PR explicitly excludes that directory from being served due to issues it caused with Jekyll trying to interpret the files to dynamically serve the content.

#### How did you catch this?

Pull in the scripts locally, install Jekyll, and run `jekyll serve --trace`. You'll see:

`Liquid Exception: Liquid error (line 13): wrong number of arguments (given 1, expected 2..3) in charts/kubecost/charts/thanos/templates/query-frontend-ingress.yml`

Adding kubecost is exactly when our builds started to fail, but I think the fix should be on our end, not for the chart to be modified.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,0 +1,1 @@
+exclude: [charts]


aiyengar2:
Probably should backport this change to `rancher/charts` + `rancher/rke2charts` and maybe add this to the unified docs at https://github.com/rancher/charts-build-scripts/tree/master/templates/template.

StrongMonkey (approved):
:100:

PennyScissors:
LGTM
Can add the trace option or some sort of logging to make any potential problems in the future easier to debug?

PennyScissors (approved):
LGTM
Can we add the trace option or some sort of logging to make any potential problems in the future easier to debug?

aiyengar2:
> LGTM
> Can we add the trace option or some sort of logging to make any potential problems in the future easier to debug?

I'm not sure if there's a way to display it on Github; they automatically try to based on [this](https://docs.github.com/en/github/working-with-github-pages/about-jekyll-build-errors-for-github-pages-sites): 
```
A "Page build failed" message means your build failed to complete. If Jekyll is able to detect a reason for the failure, you'll see a descriptive error message.
```

Easiest way to debug seems to just pull in the local copy and run the Jekyll command.

cbron (approved):

PR #79, opened 2021-03-03
Add initial version of CloudCasa Helm chart

bobadair:
This is the same chart that was added to rancher/helm3-charts in PR https://github.com/rancher/helm3-charts/pull/50, but now referenced externally.


packages/cloudcasa/overlay/questions.yaml
```diff
@@ -0,0 +1,13 @@
+namespace: cloudcasa-io
+labels:
+  io.rancher.certified: partner
+categories:
+- Application
```
PennyScissors, line 5: Read my other comment first.
If you decide to keep the files in the overlay directory here, please remove these.

PennyScissors (changes requested):
Hello @bobadair, this PR looks great overall, but still needs a few minor changes before I can approve. I'll write the steps of what's needed to make it easier for you.

1. Run `make clean` in your fork
2. Run `make prepare CHART=cloudcasa`
3. Changes needed:
    - Add the logo.png for your chart in the overlay directory (currently missing). You could also use a URL to reference a logo hosted somewhere.
    - Update the reference to the logo.png in `Chart.yaml` to `file://./logo.png` (contents of overlay are copied into the chart's root directory)
    -  Add the annotation `catalog.cattle.io/namespace: cloudcasa-io` in `Chart.yaml` to lock the namespace to `cloudcasa-io`. Locking it through the `questions.yaml` doesn't work for packages in this repository. 
    - Delete the file `Chart.yaml.orig`. It gets created to solve conflicts when applying a patch fails and should be removed once issues are resolved.
    - **Optional**: As mentioned before, the contents of overlay get copied to the chart's root directory. Since your chart has an `app-readme.md`, and `questions.yaml` upstream, they are just getting overwritten when the package is generated. Ideally, you would remove those from upstream and keep them here since they are Rancher specific and it's easier for me to review if you make any changes. However, you are currently supporting the same chart in multiple repositories, so it might be easier for you to keep them upstream and just delete them from the overlay directory here to avoid needing to update the same file in multiple places.
4. Run `make patch CHART=cloudcasa` to save your changes
5. Run `make clean' once again
6. You are ready to git add the package/cloudcasa and update this PR

bobadair:
Thanks @PennyScissors. We'll try to get these changes made by Monday. We left those files in the overlay directory (and the .patch file) because we were debating removing those customizations from the upstream chart repo. However, as you noted we are still maintaining another "Rancherized" version of the chart, so we'll have to see what ends up being easiest.

bobadair:
Sorry. We had an issue with the PR branch, so I've closed this PR and created a new one with the requested changes: #80.

PR #86, opened 2021-05-17
Add Federator.ai v4.5.1

prophetstor-ai:
Signed-off-by: CheeSiong.Lee <CheeSiong.Lee@ProphetStor.com>

### DIFF
--- a/packages/federatorai/federatorai.patch
+++ b/packages/federatorai/federatorai.patch
@@ -1,0 +1,13 @@
+diff -x '*.tgz' -x '*.lock' -uNr packages/federatorai/charts-original/Chart.yaml packages/federatorai/charts/Chart.yaml
+--- packages/federatorai/charts-original/Chart.yaml
++++ packages/federatorai/charts/Chart.yaml
+@@ -1,5 +1,9 @@
+ apiVersion: v1
+ appVersion: 4.5.1-ga
++annotations:
++  catalog.cattle.io/certified: partner
++  catalog.cattle.io/release-name: federatorai
++  catalog.cattle.io/display-name: Federator.ai
+ description: Federator.ai helps enterprises optimize cloud resources, maximize application performance, and save significant cost without excessive over-provisioning or under-provisioning of resources, meeting the service-level requirements of their applications.
+ home: https://www.prophetstor.com
+ icon: https://raw.githubusercontent.com/prophetstor-ai/public/master/images/logo.png
--- a/packages/federatorai/overlay/questions.yaml
+++ b/packages/federatorai/overlay/questions.yaml
@@ -1,0 +1,79 @@
+questions:
+#image configurations
+- variable: defaultImage
+  default: true
+  description: "Use default Federator.ai image or specify a custom one"
+  label: Use Default Federator.ai Image
+  type: boolean
+  show_subquestion_if: false
+  group: "Container Images"
+  subquestions:
+  - variable: image.repository
+    default: "quay.io/prophetstor/federatorai-operator-ubi"
+    description: "Federator.ai Operator image name"
+    type: string
+    label: Federator.ai Operator Image Name
+  - variable: image.tag
+    default: "v4.5.1-ga"
+    description: "Federator.ai Operator image tag"
+    type: string
+    label: Federator.ai Operator Image Tag
+#service configurations
+- variable: federatorai.imageLocation
+  default: "quay.io/prophetstor"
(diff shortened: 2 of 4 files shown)


packages/federatorai/overlay/questions.yaml
```diff
@@ -0,0 +1,80 @@
+rancher_min_version: 2.2.0
```
PennyScissors, line 1: Please remove this, this constraint isn't supported in this repository.

packages/federatorai/federatorai.patch
```diff
@@ -0,0 +1,18 @@
...
++  catalog.cattle.io/release-name: federatorai
+ description: Federator.ai helps enterprises optimize cloud resources, maximize application performance, and save significant cost without excessive over-provisioning or under-provisioning of resources, meeting the service-level requirements of their applications.
+ home: https://www.prophetstor.com
+ icon: https://raw.githubusercontent.com/prophetstor-ai/public/master/images/logo.png
+@@ -16,4 +19,4 @@
+ name: federatorai
+ sources:
+ - https://www.prophetstor.com
+-version: 4.5.1
++version: 4.5.1-1
```
PennyScissors, line 18: We append the packageVersion to the chart's version to generate the version of the new package that gets generated. Since you are appending -1 here, the package's version will result in 4.5.1-1<packageVersion> instead of 4.5.1<packageVersion>-1 as we expect. Is this a necessary change, or is it possible this can be reverted?

packages/federatorai/federatorai.patch
```diff
@@ -0,0 +1,18 @@
+diff -x '*.tgz' -x '*.lock' -uNr packages/federatorai/charts-original/Chart.yaml packages/federatorai/charts/Chart.yaml
+--- packages/federatorai/charts-original/Chart.yaml
++++ packages/federatorai/charts/Chart.yaml
+@@ -1,5 +1,8 @@
+ apiVersion: v1
+ appVersion: 4.5.1-ga
++annotations:
++  catalog.cattle.io/certified: partner
++  catalog.cattle.io/release-name: federatorai
```
PennyScissors, line 9: Please add the following annotation for the name you want displayed in the UI.
`catalog.cattle.io/display-name: Federator AI`
I used Federator AI as an example, but you can set it to any name you think is best.

prophetstor-ai: Fixed by forcibly commit.

packages/federatorai/overlay/questions.yaml
prophetstor-ai: Fixed by forcibly commit.

packages/federatorai/overlay/questions.yaml
```diff
@@ -0,0 +1,79 @@
...
+  - variable: federatorai.persistence.influxdb.dataStorage.size
+    default: "100Gi"
+    description: "Influxdb data volume Size"
+    type: string
+    label: Influxdb Data Volume Size
+  - variable: federatorai.persistence.fedemeterInfluxdb.dataStorage.size
+    default: "10Gi"
+    description: "Fedemeter influxdb data volume Size"
+    type: string
+    label: Fedemeter Influxdb Data Volume Size
```
PennyScissors, line 79: Just FYI, these are good, no need to add groups to these because the parent question has a group which applies to all subquestions.

packages/federatorai/overlay/questions.yaml
```diff
@@ -0,0 +1,79 @@
...
+  type: string
+  label: Federator.ai Dashboard Port
+- variable: services.rest.nodePort
+  required: true
+  default: "31011"
+  description: "The port where the Federator.ai REST listens to"
+  type: string
+  label: Federator.ai REST Port
+- variable: federatorai.persistence.enabled
+  default: true
```
PennyScissors, line 47: The chart installs corrently, and overall looks ready to merge. However, these aren't showing up in the UI because they are missing the `group: <group-name>`. Please update these and I'll approve/merge.

packages/federatorai/overlay/questions.yaml
prophetstor-ai: Added `group: <group-name>`.

packages/federatorai/overlay/questions.yaml
prophetstor-ai: Add `group: <group-name>`.

PennyScissors (approved):

PR #87, opened 2021-05-20
k8s-triliovault-operator release update v2.0.5

bhagirathhapse:
This pull request contains:

1. New helm package update for k8s-triliovault-operator v2.0.5
2. Updated readme.md file to have 80 chars in single line

### DIFF
--- a/packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
+++ b/packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
@@ -4,7 +4,7 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/k8s-triliovault-operator/charts-origin
 @@ -10,3 +10,6 @@
  sources:
  - https://github.com/trilioData/k8s-triliovault-operator
- version: v2.0.2
+ version: v2.0.5
 +annotations:
 +  catalog.cattle.io/certified: partner
 +  catalog.cattle.io/release-name: k8s-triliovault-operator
--- a/packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
+++ b/packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
@@ -1,10 +1,59 @@
 diff -x '*.tgz' -x '*.lock' -uNr packages/k8s-triliovault-operator/charts-original/Chart.yaml packages/k8s-triliovault-operator/charts/Chart.yaml
 --- packages/k8s-triliovault-operator/charts-original/Chart.yaml
 +++ packages/k8s-triliovault-operator/charts/Chart.yaml
-@@ -10,3 +10,6 @@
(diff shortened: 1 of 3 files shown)


PennyScissors (approved):

packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
```diff
@@ -4,7 +4,7 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/k8s-triliovault-operator/charts-origin
 @@ -10,3 +10,6 @@
  sources:
  - https://github.com/trilioData/k8s-triliovault-operator
- version: v2.0.2
+ version: v2.0.5
 +annotations:
 +  catalog.cattle.io/certified: partner
 +  catalog.cattle.io/release-name: k8s-triliovault-operator
```
PennyScissors, line 10: Could you please add this additional annotation for the name that will be displayed in the UI?
`catalog.cattle.io/display-name: TrilioVault Kubernetes Operator`
I used TrilioVault Kubernetes Operator as an example, but you can change it to whatever you think is best.

PennyScissors (commented):
Overall looks good to me, just need those small changes and it should be ready to merge.

packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
```diff
@@ -4,7 +4,7 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/k8s-triliovault-operator/charts-origin
 @@ -10,3 +10,6 @@
  sources:
  - https://github.com/trilioData/k8s-triliovault-operator
- version: v2.0.2
+ version: v2.0.5
```
PennyScissors, line 7: can you patch this version to `2.0.5`? We have an updating upcoming that requires valid semver, and the `v` prefix will cause problems as it is not valid semver.

packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
bhagirathhapse: Sure, I will fix this into the new commit

packages/k8s-triliovault-operator/k8s-triliovault-operator.patch
bhagirathhapse: I have fixed this while generating  the chart. I still doubt if this would fix future issues may arise due to changes in your scripts. The prefix 'v' is present in the operator code itself which is already released. I would suggest engg team to remove it altogether from the next operator release.

PennyScissors:
Closing in favor of this [PR](https://github.com/rancher/partner-charts/pull/95).

@bhagirathhapse We needed to get this merged ASAP and some changes were still needed to I cherry-picked your commits and made the changes in a different PR.

PR #90, opened 2021-06-02
Update CrowdStrike Falcon package display name

redhatrises:


redhatrises:
Ran `make validate CHART=falcon-sensor`, and it is passing locally. When I run `make validate` looks like there are failures around the recent artifactory updates but not sure if it is me or not.

```
$ make validate CHART=falcon-sensor                                                        
./scripts/bootstrap
Please make sure you have helm v3 and yq v3 installed
./scripts/validate
Chart falcon-sensor has no dependencies; skipping step to pull in dependencies
patching file Chart.yaml
Successfully packaged chart and saved it to: assets/falcon-sensor/falcon-sensor-0.9.301.tgz
```

PennyScissors:
Hello @redhatrises, you are correct, the artifactory charts are causing the validation to fail so I've contacted them to see what we can do. As for your chart, it should be okay as long as it passes validation.

PennyScissors:
Could you please add an app-readme.md (required)? You could also add a questions.yaml (optional), if you want to make configuring your chart for common cases easier. You can find more information about this in our guide [here](https://github.com/rancher/partner-charts/tree/main-source#5-add-overlay-files).

packages/falcon-sensor/package.yaml
```diff
@@ -1,2 +1,2 @@
 url: https://github.com/CrowdStrike/falcon-helm/releases/download/0.9.3/falcon-sensor-0.9.3.tgz
-packageVersion: 00
+packageVersion: 01
```
PennyScissors, line 2: It looks like this PR only adds the display-name annotation to change the chart's name in the UI. If that's the case, then it might be better to remove this. This change will generate a new package that essentially has the same underlying chart as the previous version and could be confusing for users.

PennyScissors (commented):
Hello @redhatrises, you are correct, the artifactory charts are causing the validation to fail so I've contacted them to see what we can do. As for your chart, it should be okay as long as it passes validation.

packages/falcon-sensor/package.yaml
redhatrises: That makes sense. Thanks @PennyScissors ! I have removed the package version bump.

redhatrises:
> Hello @redhatrises, you are correct, the artifactory charts are causing the validation to fail so I've contacted them to see what we can do. As for your chart, it should be okay as long as it passes validation.

Okay cool. Didn't want to assume.

Output locally from `make validate`

```
$ make validate CHART=falcon-sensor                                                                                                                                  (update_falcon_package✱)
./scripts/bootstrap
Please make sure you have helm v3 and yq v3 installed
./scripts/validate
Chart falcon-sensor has no dependencies; skipping step to pull in dependencies
patching file Chart.yaml
Successfully packaged chart and saved it to: assets/falcon-sensor/falcon-sensor-0.9.300.tgz
```

PennyScissors (approved):

PR #93, opened 2021-06-11
[Staging] Migrate to charts-build-scripts v0.2.1

aiyengar2:
Switches scripts from old Bash scripts to new Go scripts based on https://github.com/rancher/charts-build-scripts

Related Issue: https://github.com/rancher/rancher/issues/31500


packages/artifactory-ha/package.yaml
```diff
@@ -1,2 +1,2 @@
-url: https://repo.chartcenter.io/jfrog/artifactory-ha-4.13.0.tgz
+url: https://charts.jfrog.io/artifactory-ha-4.13.0.tgz
```
aiyengar2, line 1: @PennyScissors can you confirm that this is expected from the partner?

README.md
```diff
@@ -18,10 +18,13 @@ files on top of it.
     to emit the warning `manifest_sorter.go:175: info: skipping unknown hook: "crd-install"`
```
aiyengar2, line 18: @PennyScissors I manually added changes to this file and moved the README.md generated from the charts-build-scripts over to docs/README.md since you requested that we do not change the main README.md on this repository. Let me know if anything needs to be fixed here!

docs/README.md
```diff
@@ -0,0 +1,126 @@
+## Staging Branch
```
aiyengar2, line 1: Manually moved but contents were auto-generated.

packages/artifactory-jcr/package.yaml
```diff
@@ -1,2 +1,2 @@
-url: https://repo.chartcenter.io/jfrog/artifactory-jcr-3.4.0.tgz
+url: https://charts.jfrog.io/artifactory-jcr-3.4.0.tgz
```
aiyengar2, line 1: @PennyScissors can you confirm that this is expected from the partner?

packages/artifactory-jcr/generated-changes/dependencies/artifactory/dependency.yaml
```diff
@@ -0,0 +1,2 @@
+workingDir: ""
+url: https://charts.jfrog.io/artifactory/api/helm/jfrog-charts/artifactory-11.7.4.tgz
```
aiyengar2, line 2: FYI the partner will also need to manually keep this dependency file up-to-date. It does not automatically get modified on rebasing the upstream.

Does this need to be called out in the docs or announced somewhere else?

packages/cloudcasa/generated-changes/patch/Chart.yaml.patch
```diff
@@ -10,10 +9,12 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/cloudcasa/charts-original/Chart.yaml p
...
  - backup
  - Catalogic
-@@ -13,3 +13,8 @@
+@@ -12,4 +12,8 @@
+ - email: info@catalogicsoftware.com
    name: catalogicsoftware
  name: cloudcasa
- version: "1.0"
+-version: "1.0"
++version: "1.0.0"
```
aiyengar2, line 17: @PennyScissors can you confirm from the partner that this is okay? This change was required since the partner is not following proper SemVer versioning, which might / should break on a Helm install with Helm >3.5.2.

Even if it does not break, the Helm community could break it at any time (since the version is expected to be strictly SemVer, regardless of whether Helm errors on it now or now) so it is better to fix now either way.

packages/datadog/generated-changes/dependencies/kube-state-metrics/dependency.yaml
```diff
@@ -0,0 +1 @@
+url: https://charts.helm.sh/stable/packages/kube-state-metrics-2.8.11.tgz
```
aiyengar2, line 1: Same comment about needing to manually update dependency versions now.

packages/dynatrace-oneagent-operator/generated-changes/overlay/app-readme.md
```diff
@@ -1,6 +0,0 @@
-# Dynatrace OneAgent Operator
```
aiyengar2: `dynatrace-oneagent-operator` had changes stored in the patch file as well as overlaid on top, so this change was mostly the scripts adjusting to storing the minimum required diff / overlay (e.g. just a patch on the questions.yml since the upstream app-readme.md is equivalent to this).

packages/falcon-sensor/generated-changes/overlay/questions.yaml
```diff
@@ -1,97 +0,0 @@
-questions:
```
aiyengar2: removed in favor of patch

packages/falcon-sensor/generated-changes/overlay/app-readme.md
```diff
@@ -1,9 +0,0 @@
-# CrowdStrike Falcon
```
aiyengar2: overlaid app-readme.md == upstream app-readme.md in the chart that the URL points to, so this was not necessary and deleted.

packages/federatorai/generated-changes/overlay/app-readme.md
```diff
@@ -1,39 +0,0 @@
-# <img src="https://raw.githubusercontent.com/prophetstor-ai/public/master/images/logo.png" width=60/> Federator.ai Operator
```
aiyengar2: same comment re: removing unnecessary files to store minimum diff

packages/hpe-flexvolume-driver/generated-changes/patch/questions.yml.patch
```diff
@@ -0,0 +1,20 @@
+--- charts-original/questions.yml
```
aiyengar2, line 1: same comment re: removing unnecessary files to store minimum diff

packages/portshift-operator/generated-changes/patch/questions.yaml.patch
```diff
@@ -0,0 +1,11 @@
+--- charts-original/questions.yaml
```
aiyengar2, line 1: same comment re: removing unnecessary files to store minimum diff

packages/streamsets/package.yaml
```diff
@@ -1,2 +1,2 @@
-url: https://github.com/thomasganka/helm-charts/blob/master/docs/stable/control-agent-2.0.1.tgz?raw=true
+url: https://github.com/thomasganka/helm-charts/raw/master/docs/stable/control-agent-2.0.1.tgz
```
aiyengar2, line 1: @PennyScissors can you double check with the partner that this looks right?

configuration.yaml
```diff
@@ -1,7 +1,7 @@
 template: staging
 
 validate:
-- url: https://github.com/rancher/partner-charts.git
-  branch: main
+- url: https://github.com/aiyengar2/partner-charts.git
+  branch: migrate_to_v0.2.1_live
```
aiyengar2, line 5: NOTE: do not merge this PR till https://github.com/rancher/partner-charts/pull/94 is merged. This commit is a TEMPORARY commit that will be removed but is currently necessary.

aiyengar2:
@PennyScissors this PR is ready to review.

packages/artifactory-jcr/generated-changes/dependencies/artifactory/dependency.yaml
jiaqiluo: I think we can add this to the docs for charts-build-scripts v0.2.1
It has happened once for Fleet that the dependency was not changed to the latest when the Fleet itself was changed to the latest.

packages/artifactory-ha/package.yaml
PennyScissors: I can't confirm for them, but this will fix our validation which is failing because the old URL gets a 405 Not Allowed, whereas the new pulls the tarball correctly. Seems to be the same chart and I notified them their chart had problems before so I think it should be okay to keep this.

packages/artifactory-jcr/package.yaml
PennyScissors: > I can't confirm for them, but this will fix our validation which is failing because the old URL gets a 405 Not Allowed, whereas the new pulls the tarball correctly. Seems to be the same chart and I notified them their chart had problems before so I think it should be okay to keep this.

Same thing here. Is basically the same chart, but one doesn't require a license.

packages/artifactory-jcr/generated-changes/dependencies/artifactory/dependency.yaml
PennyScissors: I will add it to the docs as part of the steps for upgrading.

jiaqiluo (approved):
LGTM. 
But I wonder why we have those following 2 empty files? 
```
packages/federatorai/generated-changes/patch/requirements.patch 
packages/k8s-triliovault-operator/generated-changes/overlay/app-readme.md
```

Update: 
Ignore this comment.  I was looking at each comment which gave me the wrong impression. If I look at the list of final changes, the 1st file is removed and 2nd is not empty.

packages/streamsets/package.yaml
PennyScissors: confirmed, this one is good.

packages/artifactory-jcr/package.yaml
PennyScissors: Confimed with partner this is fine.

packages/artifactory-ha/package.yaml
PennyScissors: Confimed with partner this is fine.

PennyScissors (approved):

PR #103, opened 2021-06-25
Update README

PennyScissors:


aiyengar2 (commented):
Should we add something about the +up annotation too?

PennyScissors:
> Should we add something about the +up annotation too?

Updated to include information on the +up annotation and removed the `v` prefix from examples since it looks like we remove that when a package gets generated.

aiyengar2 (approved):
LGTM

PR #109, opened 2021-07-05
Curiefense updated PR

JuanPabloGarciaMonzon:
In this new PR I have made only 2 commits as the contribution rules demands, 1 with the "make prepare" and "make patch" commands and the sencond one with the "make charts" and "make clean" commands. Also the "make validate" command has been run in my local machine and it has deployed a successful response.


PennyScissors:
Hello @JuanPabloGarciaMonzon, the structure of the PR looks correct, but your first commit is incorrect. You should be adding a package as described in our instructions [here](https://github.com/rancher/partner-charts/tree/main-source#2-set-up-your-package-to-track-an-upstream-chart-skip-if-upgrading-existing-package), and not a standard helm chart. You should also make sure to add the annotations and other requirements we have listed in that same document. Thanks

JuanPabloGarciaMonzon:
ok I'll fix ii right away, although the package I remember creating it with mkdir not with helm create, but anyways I'll check it out. Thanks for the review.

PennyScissors:
No problem, what I mean by standard chart is that it has all the contents of your charts. A package doesn't not have those files, instead it has information stored within the package.yaml in order to pull your chart from upstream (1st commit), to apply the rancher specific things on top of it and generate a new chart (This is what the 2nd commit `make charts` is for).

JuanPabloGarciaMonzon:
I'm going to create a new PR in order to have the commits organized like this PR, i think that the error that i made was to not run make clean after the make patch in the first commit. I read the MD again and that is the only step that i remember not doing.

JuanPabloGarciaMonzon:
I am creating a new PR with the updated Commits with the changes i have made.

PR #110, opened 2021-07-08
TrilioVault for Kubernetes Operator update v2.1.0

bhagirathhapse:
This PR contains:
- Update of k8s-triliovault-operator v2.1.0
- Updated patch file due to removal of 'v' from the Chart.yaml

### DIFF
--- a/packages/k8s-triliovault-operator/generated-changes/patch/Chart.yaml.patch
+++ b/packages/k8s-triliovault-operator/generated-changes/patch/Chart.yaml.patch
@@ -1,20 +1,9 @@
 --- charts-original/Chart.yaml
 +++ charts/Chart.yaml
-@@ -1,6 +1,7 @@
- apiVersion: v1
- appVersion: v2.0.5
--description: K8s-TrilioVault-Operator is an operator designed to manage the K8s-TrilioVault Application Lifecycle.
-+description: K8s-TrilioVault-Operator is an operator designed to manage the K8s-TrilioVault
-+  Application Lifecycle.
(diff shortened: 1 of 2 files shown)


PennyScissors:
Hello @bhagirathhapse, can you please refactor this PR to have 2 commits? 1st commit should be the changes to the chart to update it to v2.1.0 (for example, the change to update the url in package.yaml and other things). The 2nd commit should be committing the results of running `make charts`. Also, I see some things missing from your patch that were added in your previous update (for example, [this](https://github.com/rancher/partner-charts/blob/6dfcbb04bdf8e3906cb5b43e73362a942a12b33a/packages/k8s-triliovault-operator/generated-changes/patch/Chart.yaml.patch#L21) annotation), so you might need to rebase.

PennyScissors (changes requested):
Hello @bhagirathhapse, can you please refactor this PR to have 2 commits as explained [here](https://github.com/rancher/partner-charts/tree/main-source#8-pull-request)? 1st commit should be the changes to the chart to update it to v2.1.0 (for example, the change to update the url in package.yaml and other things). The 2nd commit should be committing the results of running `make charts`. Also, I see some things missing from your patch that were added in your previous update (for example, [this](https://github.com/rancher/partner-charts/blob/6dfcbb04bdf8e3906cb5b43e73362a942a12b33a/packages/k8s-triliovault-operator/generated-changes/patch/Chart.yaml.patch#L21) annotation), so you might need to rebase your fork.

bhagirathhapse:
@PennyScissors Let me remove this PR and create a new one since we require only 2 commits.

PR #111, opened 2021-07-14
Neuvector: Updated with 1.8.0 version

selvamt94:


PennyScissors (changes requested):
Hello @selvamt94, the structure of the PR looks correct, but your first commit is incorrect. You should be adding a package as described in our instructions [here](https://github.com/rancher/partner-charts/tree/main-source#2-set-up-your-package-to-track-an-upstream-chart-skip-if-upgrading-existing-package), and not a standard helm chart. You should also make sure to add the annotations and other requirements we have listed in that same document (for example, you are missing the display-name annotation and kubeVersion in Chart.yaml). Thanks.

PR #115, opened 2021-07-20
Kubecost upgrade 1.83.1

AjayTripathy:
Bump Kubecost to 1.83.1

Note that we have not applied a kubeVersion field.  Forcing semver on our users probably won't work. GKE doesn't respect semver in listing its versions of kubernetes. 

Even if you allow prereleases (https://github.com/helm/helm/issues/3810), the semver on GKE doesn't seem to work for us:

```
atripathy@Ajays-MacBook-Pro cost-analyzer-helm-chart % helm upgrade kubecost -n kubecost ./cost-analyzer

Error: UPGRADE FAILED: chart requires kubeVersion: “>=1.8.0-0” which is incompatible with Kubernetes v1.18.17-gke.1901
the .1901 at the edge of google's version here is messing up the helm semver checker.
```

We install kubecost across a wide range of k8s providers who don't necessarily respect semver when versioning kubernetes.


PennyScissors (changes requested):
Hey @AjayTripathy, your first 2 commits adding the package is removing files that it shouldn't remove (like Chart.yaml.patch, app-readme.md, and questions.yaml), and adding others that it shouldn't add (like .orig and .rej files which are created for rebasing when a patch doesn't apply cleanly after a make prepare). We also require some annotations that are missing from your package since the patch files were removed.

AjayTripathy:
Ok, thanks for that feedback. I thought I followed the steps in the directions... any ideas or pointers on what I might have done wrong? In this situation should I just be closing this PR and trying to open a new one?

PennyScissors:
Hey @AjayTripathy, I'm not really sure what could've gone wrong, but you can reach out to me on the #rancher-kubecost Slack channel or directly, and we can figure it out together.

AjayTripathy:
Hi @PennyScissors could you take a look at https://github.com/rancher/partner-charts/pull/140 ? I've had to hand this off-- closing this PR for now.

PR #126, opened 2021-07-28
(Trilio) TrilioVault for Kubernetes release update v2.1.0 

bhagirathhapse:
This PR consists of:
- URL change in package.yaml for 2.1.0 release update
- Annotations for partner, package-name and display-name in Chart.yaml
- Newly generated Chart.yaml.patch file
- Newly generate assets/ charts/ directories.


PennyScissors (changes requested):
Hello @bhagirathhapse, we now require setting a kubeVersion in Chart.yaml as explained [here](https://github.com/rancher/partner-charts/tree/main-source#set-the-kubeversion-and-annotations-for-new-packages-only). Also, I'm running into some issues with the service account when deploying the chart as I mentioned in my other comment. Please make sure to test the chart in Rancher.

charts/k8s-triliovault-operator/k8s-triliovault-operator/2.1.0+upv2.1.0/templates/serviceAccount.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: k8s-triliovault-operator
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: k8s-triliovault-operator
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
```
PennyScissors, line 8: I'm running into the following problem when installing the chart which seems to be caused by this line. It uses the whole version including the build metadata with the `+` which is not a valid character for a label and fails.

```
Error: ServiceAccount "k8s-triliovault-operator" is invalid: metadata.labels: Invalid value: "k8s-triliovault-operator-2.1.0+upv2.1.0": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue', or 'my_value', or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

Resolved service account template taken from the output of `helm template assets/k8s-triliovault-operator/k8s-triliovault-operator-2.1.0+upv2.1.0.tgz`
```
apiVersion: v1                                                                                                                      
kind: ServiceAccount                                                                                                                
metadata:                                                                                                                           
  name: k8s-triliovault-operator                                                                                                    
  namespace: default                                                                                                                
  labels:                                                                                                                           
    app: k8s-triliovault-operator                                                                                                   
    chart: "k8s-triliovault-operator-2.1.0+upv2.1.0"                                                                                
    release: "RELEASE-NAME"                                                                                                         
    heritage: "Helm" 
```

PennyScissors (changes requested):

charts/k8s-triliovault-operator/k8s-triliovault-operator/2.1.0+upv2.1.0/templates/serviceAccount.yaml
bhagirathhapse: @PennyScissors I understand here that the service account issue is due to the label *chart: "k8s-triliovault-operator-2.1.0+upv2.1.0"*. But originally this label was *chart: "k8s-triliovault-operator-2.1.0* and after rancher chart generation it is updated to add *+upv2.1.0*.
So even if we update the helm chart and remove 'v' from the *version* and *appVersion* parameter values, the rancher chart script will still concatenate the *appVersion* value to *version* and this problem will arise.

bhagirathhapse:
I have updated the *kubeVersion* parameter as you requested but the service account would still fail due to issue explained in above comment.

PennyScissors:
@bhagirathhapse Your last commit should be the one that adds the results of running `make charts` for validation to pass. Please drop your previous make charts commit (fce486d), and create a new commit after the kubeVersion change you submitted.

bhagirathhapse:
@PennyScissors This was my bad, some issue with access token and changes from *make charts* were not pushed. Its done now. Please check and let me know if *+* sign could still be an issue.

PennyScissors (approved):

PR #127, opened 2021-08-02
kong

CAcquaviva:


PennyScissors (changes requested):
Hello @CAcquaviva, this PR is adding a standard Helm chart, and it should be adding a package instead. Please refer to our guide [here](https://github.com/rancher/partner-charts/tree/main-source#readme) that explains the new package based process and what's expected in your PR. Thanks

PR #129, opened 2021-08-04
(Nutanix) update upstream to v2.3.11

tuxtof:


PennyScissors (changes requested):
Looks good overall, tested deployment in Rancher and it works fine, but the `catalog.cattle.io/display-name` annotation for what you want in the UI to be displayed is missing from Chart.yaml (more info [here](https://github.com/rancher/partner-charts/tree/main-source#4-make-changes)). Make sure your last commit is a `make charts` commits, so you will need to drop the current `make charts` commit, make the requested changes (using the make commands, do not modify the chart that gets generated directly), and then commit another `make charts` commits after.

Also, I noticed you made all Rancher specific changes upstream rather than in a patch in our repository. Was that intentional?

alex-isv (changes requested):
@tuxtof 
Missing annotation: catalog.cattle.io/display-name
-> https://github.com/rancher/partner-charts/tree/main-source#4-make-changes
Also, we don't use +upxxx versioning anymore like in your case

image

it should be 2.3.1101.tgz without +upxxx

PennyScissors:
Closing this as there hasn't been any activity for a while and we recently updated our packaging scripts and it will likely be easier to submit a new PR to avoid rebasing issues.

PR #132, opened 2021-08-10
Kong Helm Charts 3

CAcquaviva:


PennyScissors (changes requested):
Hello @CAcquaviva, the structure of this PR looks correct, and you've included some of the things we require, but the package is missing a `kubeVersion` and the annotations stated in [this step](https://github.com/rancher/partner-charts/tree/main-source#4-make-changes).

PR #134, opened 2021-08-12
(Prophetstor) Update Federator.ai to version v4.6.1 

prophetstor-ai:


alex-isv (changes requested):
We don't add +upvX.X.X to the version anymore.
You have federatorai-4.6.100+up4.6.1.tgz 
So, you need to delete the last commit, re-run make charts and commit it again.
The resulting chart should be something like federatorai-4.6.100.tgz (without +up4.6.1)

prophetstor-ai:
> We don't add +upvX.X.X to the version anymore.
> You have federatorai-4.6.100+up4.6.1.tgz
> So, you need to delete the last commit, re-run make charts and commit it again.
> The resulting chart should be something like federatorai-4.6.100.tgz (without +up4.6.1)

Done in commit the new codes.

PR #137, opened 2021-08-17
Regenerate all assets and charts from scratch

aiyengar2:
Removes all the charts / assets that were never released from the main-source branch and removes any corresponding entries in the index.yaml that were left behind on revisions.

Basic process is just the following:
1. Run `rm -rf charts assets index.yaml` and run `git checkout -- assets/README.md charts/README.md`. Commit changes.
2. Run `make validate` once; this will effectively run `make charts` and then fail. Commit changes.
3. Run `make validate` again; this will pull in the current contents of `main` and then fail. Commit changes.

Workflow is passing so this branch now passes `make validate` on the third try with the committed changes.


index.yaml
```diff
@@ -7,9 +7,9 @@ entries:
       catalog.cattle.io/release-name: ambassador
     apiVersion: v1
     appVersion: 1.13.8
-    created: "2021-06-25T13:41:52.528853-07:00"
+    created: "2021-08-16T16:58:12.196947-07:00"
     description: A Helm chart for Datawire Ambassador
-    digest: 395a8bfddf57116dabaa5f96ef027e05094b25c720947223f9167843cf4e7319
```
aiyengar2, line 12: digests do not get updated by Helm if the chart version remains the same since Helm does not update the index.yaml for existing entries.

index.yaml
```diff
@@ -33,52 +33,20 @@ entries:
     urls:
     - assets/ambassador/ambassador-6.7.1100+up6.7.11.tgz
     version: 6.7.1100+up6.7.11
-  - annotations:
```
aiyengar2, line 36: Charts like this that were added but never released are dropped from the annotations since they do not need to exist

PennyScissors (approved):

PR #140, opened 2021-08-19
(Kubecost) Upgrade Kubecost to 1.83.2

elenalape:
Attempting to fix #115.

- Commit 1: Update the upstream URL and change version — I'm not sure about the `packageVersion` numbers here, should it be 1, 2, or 0, if a new URL containing a newer version is used?
- Commit 2: output from `make charts`

As per the previous PR,

> Note that we have not applied a kubeVersion field. Forcing semver on our users probably won't work. GKE doesn't respect semver in listing its versions of kubernetes.

cc @AjayTripathy


AjayTripathy:
Hi @PennyScissors just wanted to point you to this PR and let you know @elenalape is picking this up for me. Thanks all!

packages/kubecost/generated-changes/patch/Chart.yaml.patch
```diff
@@ -1,10 +1,18 @@
 --- charts-original/Chart.yaml
 +++ charts/Chart.yaml
-@@ -4,3 +4,7 @@
+@@ -1,10 +1,14 @@
+ annotations:
++  catalog.cattle.io/certified: partner
++  catalog.cattle.io/release-name: kubecost
```
PennyScissors, line 6: The `catalog.cattle.io/display-name` annotation for what you want the UI to display is missing (more info on it [here](https://github.com/rancher/partner-charts/tree/main-source#4-make-changes)). Make sure you use the make commands and follow the workflow order to make the change, and re-run make charts again. You need to re-run make charts every time you make changes, and it should replace your previous make charts commit.

PennyScissors (changes requested):

packages/kubecost/generated-changes/patch/Chart.yaml.patch
elenalape: Thanks! I've added the display name in commit #3 and re-ran `make charts` in #4.

elenalape:
Thanks for the corrections @PennyScissors! I've made the amendments, `make validate` seems to have worked, please let me know if something doesn't look right.

elenalape:
Note: I've updated the upstream to the newest version of Kubecost which is 1.83.2

PennyScissors:
Hello @elenalape, could you please refactor this PR into 2 commits to make it easier to review? The upstream reference was updated twice and the charts were generated twice as well since there are 2 make charts commits. Your first commit should be adding the package in packages/ directory and all the requirements, and the second commit should be the output of `make charts`. When a change needs to be made, the workflow is to drop the make charts commit, amend the first commit with the changes and create a new make charts commit and this way the PR remains with 2 commits.

packages/kubecost/generated-changes/overlay/questions.yml
```diff
@@ -1,160 +0,0 @@
-questions:
```
elenalape, line 1: this file is now `.yaml` for consistency

elenalape:
Thanks @PennyScissors, makes sense. Apologies for the back and forth, hopefully, we're almost there!

(Please ignore the opening/reopening — I was thinking about renaming the branch, but that would mean another PR.)

PennyScissors:
No worries @elenalape, we can continue reusing this PR or create a new one. Whatever makes it easier for you works for me 😄. All I need is just keeping it at 2 commits if possible since it can be challenging to review when it goes past that.

The first commit includes the generated `packages/cost-analyzer/charts` directory which is generated from the `make prepare` step and is only meant to be used for making changes that will be saved to the patch files. After doing `make patch` to save your changes, you should do `make clean` which will get rid of all the unwanted generated stuff and then it should be good to commit everything in `packages/cost-analyzer`. Here is a good [PR](https://github.com/rancher/partner-charts/pull/128) for reference.

elenalape:
@PennyScissors that PR's been helpful, cheers!. Fourth time's a charm 🤞🏻

PR #141, opened 2021-08-20
Add nginx ingress support

ciarams87:
This commit ports support for the nginx-ingress-controller from the previous helm3 repo.

Tested in the latest v2.5 Rancher UI.


alex-isv (dismissed):

PennyScissors (changes requested):
The package you added in this PR is correct, but you included the `charts` directory located in your package which is incorrect. That directory is used to apply changes to the upstream chart and then save those changes in the patch files inside `generated-changes`. Once you've made the changes and ran `make patch` to save them, the next step is to run `make clean` to remove what gets generated and then commit the contents of your package. 

To fix this you can rebase to your first commit, run `make clean`, and amend your first commit with those changes. Only `generated-changes` and `package.yaml` should be present in your package directory.

ciarams87:
> The package you added in this PR is correct, but you included the `charts` directory located in your package which is incorrect. That directory is used to apply changes to the upstream chart and then save those changes in the patch files inside `generated-changes`. Once you've made the changes and ran `make patch` to save them, the next step is to run `make clean` to remove what gets generated and then commit the contents of your package.
> 
> To fix this you can rebase to your first commit, run `make clean`, and amend your first commit with those changes. Only `generated-changes` and `package.yaml` should be present in your package directory.

@PennyScissors Deleted the extra directory, thanks! I think I committed the first commit before I ran 'make clean' so this was missed.

alex-isv (approved):

PennyScissors (approved):
lgtm

PR #156, opened 2021-09-03
Fix Streamsets chart

PennyScissors:
Add missing annotation and kubeVersion


aiyengar2 (approved):
LGTM

charts/streamsets/control-agent/2.0.101/Chart.yaml
```diff
@@ -9,6 +10,7 @@ keywords:
 - streamsets
 - sdc
 - sch
+kubeVersion: 1.18 - 1.21
```
aiyengar2, line 13: Based on Helm Docs this is valid: https://v3.helm.sh/docs/topics/charts/#the-kubeversion-field

`hyphen ranges for closed intervals, where 1.1 - 2.3.4 is equivalent to >= 1.1 <= 2.3.4.`

charts/streamsets/control-agent/2.0.101/Chart.yaml
```diff
@@ -9,6 +10,7 @@ keywords:
 - streamsets
 - sdc
 - sch
+kubeVersion: 1.17 - 1.21
```
aiyengar2, line 13: index.yaml should be modified https://github.com/rancher/partner-charts/blob/6f3b252e5c217636bcd575627fb7aca0b3ac4abd/index.yaml#L456-L477

Can you delete the above entry and run `make charts` to regenerate it?

aiyengar2 (changes requested):

charts/streamsets/control-agent/2.0.101/Chart.yaml
PennyScissors: Done

aiyengar2 (approved):
LGTM

PR #162, opened 2021-09-08
Update HAProxy chart to 1.16.3

andrewheberle:
This PR updates the HAProxy ingress controller helm chart to the latest version currently released.


alex-isv (approved):
looks good

packages/haproxy/generated-changes/patch/Chart.yaml.patch
```diff
@@ -1,14 +1,9 @@
 --- charts-original/Chart.yaml
 +++ charts/Chart.yaml
-@@ -14,7 +14,11 @@
-   name: Baptiste Assmann
- - email: dkorunic@haproxy.com
-   name: Dinko Korunic
--name: kubernetes-ingress
-+name: haproxy
```
PennyScissors, line 8: Everything in this PR is correct related to the package itself is correct, but you didn't change the name back to `haproxy` when you updated the chart and this resulted in a new kubernetes-ingress chart along with the old haproxy chart. You need to rebase your first commit to add this change (Using the make commands, make prepare, apply the change, make patch, make clean, then amend commit). Make sure you delete your `make charts` commit (2nd commit), re-run it and re-commit the output so the index.yaml is updated correctly.

image

CC @alex-isv

PennyScissors (changes requested):

packages/haproxy/generated-changes/patch/Chart.yaml.patch
andrewheberle: Sorry about that.

I've rebased and re-done the change and then force-pushed to my branch so that should be sorted out now.

alex-isv:
@andrewheberle 
I still see the same names (old versions and a new version)
image

image

and 

image

andrewheberle:
@alex-isv I am not seeing the same behavior in my test Rancher deployment. I only see the one chart:

image

With the following versions:

image

andrewheberle:
Is there any further I can do on this PR so it is possible to merge?

I can’t reproduce the duplicate chart that @alex-isv is seeing.

alex-isv (approved):

PR #166, opened 2021-09-13
Adding jfrog-platform charts

vinayagg:
JFrog put all products (Artifactory, Xray, Distribution, Pipelines, Mission Control etc.) in a single distribution called JFrog-platform.
These helm charts are for JFrog platform. 
Currently they install only Artifactory and Xray.
In future, we can enable other products too.


alex-isv (changes requested):
Hi @vinayagg

You need to add annotation in your chart.yaml file:
 catalog.cattle.io/display-name: Your Chart Name Here

Please clarify the installation prereq. on a single local node as I"m getting this error:

image

I don't see an option to change to NodePort instead of LoadBalancer for services in your yaml file.

image


Is that really should be unchecked for installation >>
image

vinayagg:
Hello @alex-isv ,
1. I will make the modification. Can I do additional commit or should I rollback both commits and recreate 2 commits?
2. We do not support NodePort for services. This is HA setup. Is NodePort support required?
3. Yes, user needs to uncheck. Our OOTB charts have these on by default. I am no able to override using rancher's override mechanism. This is due to a bug in rancher, I could not find a workaround. Bug is here: https://github.com/rancher/rancher/issues/13842

Once you let me know, I will fix these. Thank you.

alex-isv:
> Hello @alex-isv ,
> 
>     1. I will make the modification to add annotation. Can I do additional commit or should I rollback both commits and recreate 2 commits?
> 
>     2. We do not support NodePort for services. This is HA setup. Is NodePort support required?
> 
>     3. Yes, user needs to uncheck. Our OOTB charts have these on by default. I am no able to override using rancher's override mechanism. This is due to a bug in rancher, I could not find a workaround. Bug is here: [Setting boolean flag in questions.yaml, show wrong value in answers rancher#13842](https://github.com/rancher/rancher/issues/13842)
> 
> 
> Once you let me know, I will fix these. Thank you.



Hi @vinayagg 

Just roll back to fix. You need to have only 2 commits at the end.

Usually NodePort option is used to test locally to replace LoadBalancer option for the service.
for ex. > 

image

which you define in your questions.yaml
-> https://github.com/alex-isv/partner-charts/tree/main-source#questions-example

image


or from the yaml config file

image

I can see it in your questions.yaml file >
image

I just don't know why this option is not visible to select from the Rancher side.

I'm not testing your product, I'm just testing deployment on the Rancher locally. So, I need to know how to make a simple configuration to be able to deploy it on a single local Rancher node.
Please test it on a local node and give me specific steps on how to do it.

vinayagg:
@alex-isv I investigated the issue. I found this is same or similar bug I hit earlier i.e. based on helm chart values turn UI elements on/off.
In this case, our questions.yaml looks as follows

`
- variable: artifactory-ha.ingress.enabled
  default: false
  description: "Expose app using Layer 7 Load Balancer - ingress"
  type: boolean
  label: Expose app using Layer 7 Load Balancer
  show_subquestion_if: true
  group: "Services and Load Balancing"
  required: true
  subquestions:
- variable: artifactory-ha.nginx.service.type
  default: "LoadBalancer"
  description: "Nginx service type"
  type: enum
  required: true
  label: Nginx Service Type
  show_if: "artifactory-ha.nginx.enabled=true&&artifactory-ha.ingress.enabled=false"
  group: "Services and Load Balancing"
  options:
  - "ClusterIP"
  - "NodePort"
  - "LoadBalancer"
`

Note first variable name: artifactory-ha.ingress.enabled
We only want to show the options if ingress is not enable. But that check is not working.
I think we will hit this issue more and more, can we escalate and get this bug fixed so we do not have to find usability impacting workarounds?

alex-isv (changes requested):
@vinayagg 
If you have any issues, you need to open a bug -> https://github.com/rancher/rancher/issues/new/choose
and describe it from your side.

From my side I should be able to deploy your chart from Rancher. Preferably from a single node.
If you have a specific steps on how to do it, please describe, so I can verify.

vinayagg:
Hello, these are previously known issues with existing jira items here. So I did not file new issues. Should I?
https://github.com/rancher/rancher/issues/34793
https://github.com/rancher/rancher/issues/13842

PR #164, opened 2021-09-12
citrix-adc-istio-ingress-gateway to version 1.11.0

ajeetas:

PR #167, opened 2021-09-15
citrix-adc-istio-ingress-gateway to version 1.11.0 

ajeetas:


ajeetas:
ingress2-rancher3
ingress2-rancher2
ingress2-rancher1
Please review changes.
Thanks,
Ajeeta

alex-isv (commented):
Updated from #164 

Prerequisite:

- enable/install istio service mesh.

There are two ways to deploy this chart

    When ingress-gateway **ADC is CPX**
        Need to set arguments:
            - Set citrixCPX to true in Deployment settings Deployment Settings
           - Accept EULA license to YES in ingressGateway Settings
            - Set Ingress gateway nodePortrequired to true in ingressGateway Settings
When ingress-gateway **ADC is VPX**

    Need to set arguments:
        - Set citrixCPX to false in Deployment settings Deployment Settings
        - Accept EULA license to YES in ingressGateway Settings
        - Set Ingress gateway netscalerUrl with ADC nsip and vserverIP with VIP of ADC in ingressGateway Settings
=============

alex-isv (approved):
Verified.

packages/citrix-adc-istio-ingress-gateway/generated-changes/patch/Chart.yaml.patch
```diff
@@ -1,9 +1,25 @@
 --- charts-original/Chart.yaml
 +++ charts/Chart.yaml
-@@ -13,3 +13,6 @@
- sources:
- - https://github.com/citrix/citrix-istio-adaptor
- version: 1.2.1
+@@ -1,15 +1,21 @@
 +annotations:
 +  catalog.cattle.io/certified: partner
++  catalog.cattle.io/display-name: citrix-adc-istio-ingress-gateway
```
PennyScissors, line 6: The display-name is the name displayed for the chart in the UI, so it should be capitalized and space separated rather than hyphenated. For example, `Citrix ADC Istio Ingress Gateway`

CC @alex-isv

PennyScissors (changes requested):
This PR contains a lot of changes saved in the patch files, which is okay if they are Rancher specific and can't be done in the upstream chart, but be aware that it will make future maintenance of this chart difficult. This is because when you update the reference of the upstream chart to the newer one, the patch files will likely fail and all the changes you have in those patch files will need to be applied manually. I would recommend making any non Rancher specific changes upstream if possible, but if you prefer to keep them here let me know and I'll approve once the other change I requested is done.

ajeetas:
Hi @PennyScissors,
These patch files are created, when we have to expose multiple applications, which we are defining using struct `SecretVolumes`. In this SecretVoulme we need to define subsequent fields like secret, volume name, mount path.
    secretVolumes:
    - name: 
      secretName: 
      mountPath: 

Like below in deployment file we are range(looping) over multiple application:
 {{- range .Values.ingressGateway.secretVolumes }}
      - name: {{ .name }}
        secret:
          secretName: {{ .secretName | quote }}
          optional: true
     {{- end }}

So I have added some extra conditions in this case.
Here I didn’t find, how we can define structure like `SecretVolumes` in question.yml to add dynamically ranging over it to expose multiple applications?

Is there way to define range over struct type variable in question .yml ? So I don’t need to expose it question.yml something like below for all subsequent field of Secretvolume:
- variable: ingressGateway.secretVolumes[0].name
  required: true
  type: string
  show_if: "ingressGateway.exposeMutipleApps=true"
  label: ingressGateway secretVolumes name

Here it's fixed like for single app will expose, but if we want to expose more than one app.

Thanks,
Ajeeta

PR #174, opened 2021-09-24
update nutanix-csi-storage to  2.4.1

tuxtof:


alex-isv:
@tuxtof 

Everything was installed for me, except its hanging on:

image
 
and some warning message

image

Can you clarify specific settings on Rancher for csi-node-ntnx-plugin

tuxtof:
Hello @alex-isv 
Warning is normal because we don't yet support CRD stable

for the DS in progress can you give me more info on why it's blocking ? describe ? events ?
and more info on your test env and i will try to reproduce it

alex-isv:
> Hello @alex-isv Warning is normal because we don't yet support CRD stable
> 
> for the DS in progress can you give me more info on why it's blocking ? describe ? events ? and more info on your test env and i will try to reproduce it

@tuxtof 
I used just a default k3s  local rancher 2.6 instance. I didn't change any default helm settings.
Although, it's working fine when tested on rke cluster.
Do you support k3s from your side?

tuxtof:
never test it with the local k3s or with k3s at all
let me check and i come back to you

alex-isv (changes requested):
If you don't support k3s from your side, you can just add a comment in your readme file on the prereq.

tuxtof:
Plan to test it this week

tuxtof:
@alex-isv 
just made the test today with the k3s local instance , and the reason it failed was the lack if iscsi prerequisite
i also test some other CSI chart like the HPE against k3s and result is the same

how do you want we handle the situation
no think CSI is very relevant in the case of the local k3s cluster no ???

alex-isv:
> @alex-isv just made the test today with the k3s local instance , and the reason it failed was the lack if iscsi prerequisite i also test some other CSI chart like the HPE against k3s and result is the same, it don't work
> 
> how do you want we handle the situation no think CSI is very relevant in the case of the local k3s cluster no ???

@tuxtof 
Like I said, if only rke is supported from your side, just provide a detailed description in your Readme file on the installation steps and point that k3s is not supported.
 If you have any prereq. to make it work on k3s, just provide installation steps with screenshots.
From my side, I'm not testing the functionality of your chart. I just need to be able to deploy it from the Rancher side. I'm testing charts for both; rke and k3s and if one is not supported for some reason, just list it somewhere in the Readme prereq.
Other than that everything looks good to me.

tuxtof:
prerequisite are already in the readme, it is not specific to k3s

tuxtof:
@alex-isv 
i add specific info for k3s in description hope it is ok for you

alex-isv:
@tuxtof 
You need to have only 2 commits in your PR. Clean the old ones.

PR #144, opened 2021-08-25
OpenEBS: Updated with 2.11.0 latest chart

is-ashish:
This PR contains changes

- URL changed in package.yaml for 2.12.0
- Annotations for partner, package-name in Chart.yaml
- Newly generated Chart.yaml.patch file
- Newly generate assets/, charts/ directories.

PR #175, opened 2021-09-24
OpenEBS: Updated with 3.0.0 latest chart

is-ashish:
This PR contains changes

- URL changed in package.yaml for 3.0.0
- Annotations for partner, package-name in Chart.yaml
- Newly generated Chart.yaml.patch file
- Newly generate assets/, charts/ directories.


alex-isv:
@is-ashish 
Is PR https://github.com/rancher/partner-charts/pull/144
should be removed and replaced with this one?

is-ashish:
hi @alex-isv 
we need both PR that will give option to select specific release OpenEBS 2.12 as well as OpenEBS 3.0.0.

First should be merged [PR-144](https://github.com/rancher/partner-charts/pull/144).https://github.com/rancher/partner-charts/pull/144

Thanks

PennyScissors:
Hello @is-ashish, we don't normally release multiple versions of the same chart, just the latest for each chart that's in the main-source branch. Unless you need both charts released for some reason we are not aware of, I would recommend you close the other PR and keep this one open. Otherwise, you will need to rebase/refactor this PR completely if we merge the other one because you are modifying the same chart from two different PRs. The UI will still show the latest as the default option for users to upgrade to so the user would need to select 2.12 specifically to upgrade to it.

is-ashish:
Have merged this PR changes into [PR-144](https://github.com/rancher/partner-charts/pull/144). So it closing this PR.

PR #178, opened 2021-10-03
InAccel FPGA Operator chart

eliaskoromilas:
We would like to have this chart in the **Cluster Tools** catalog (since FPGA provisioning concerns cluster admins), but I couldn't get the `catalog.cattle.io/type: cluster-tool` annotation to work. Does this annotation apply only to charts in the two in-tree Helm repositories (**Partner**/**Rancher**)? Is this why it didn't work with my 3rd-party test repo/chart?

I noticed that the `keywords` list is used to categorize the Helm charts. Is there a plan to introduce an annotation (e.g. `catalog.cattle.io/categories: comma,separated,categories`) for this, too?

Our plan is to pull these changes up to the original chart.


alex-isv (changes requested):
Hi @eliaskoromilas 
I have no idea about this annotation: cluster-tool. I'll check internally on that.
Also, do you support k3s for your chart, because it's not even visible to deploy if using k3s as a test instance. (if not, you need to mention it in your readme file)
From rke cluster, at least I can see your chart, but when deploying, I'm getting the following error ->
image

You need to get rid of +rancher to make it work as + is not valid here.

eliaskoromilas:
Hi @alex-isv, thanks for testing this.

Regarding the chart version, to be honest it felt like a natural choice for a downstream release. The Helm linter did not complain but it seams that `kubelet` does with the generated labels. I omitted `+rancher` in the version string. It should be fine now.

Our chart supports every K8s distribution including K3s. The reason you can't see the chart in the local cluster is the `catalog.cattle.io/scope: downstream` annotation. The control plane is not meant to host FPGAs nor FPGA-accelerated apps, so enabling our **Cluster Tool** only in user clusters can be considered an architecture recommendation (https://rancher.com/docs/rancher/v2.6/en/overview/architecture-recommendations/#separation-of-rancher-and-user-clusters).

The following are all the annotations that we have included, but since there is limited documentation around them, could you please verify their validity (deprecation roadmap, etc.)? 

```yaml
catalog.cattle.io/display-name: InAccel FPGA Operator
catalog.cattle.io/namespace: kube-system
catalog.cattle.io/os: linux
catalog.cattle.io/release-name: inaccel
catalog.cattle.io/scope: downstream
catalog.cattle.io/type: cluster-tool
```

Also, what about the chart categories? Would it make sense to introduce a related annotation?

**Note**: This chart is obviously dependent on the cluster having suitable FPGA hardware. Cloud FPGA instances:
* AWS - F1 Family
* Azure - NP Series

alex-isv (changes requested):
@eliaskoromilas 
Your chart version is still listing +rancher
image

As a result, it's failing during deployment.
image

You need to fix a template in your chart to use something else.
As for your annotation question.
We require only 3 annotations from our side;
https://github.com/rancher/partner-charts/tree/main-source#set-the-kubeversion-and-annotations-for-new-packages-only
.....
  annotations:
  catalog.cattle.io/certified: partner # Enables the "partner" badge in the UI for easier identification
  catalog.cattle.io/release-name: chart-name-here # Your chart's name in kebab-case, this is used for deployment
  catalog.cattle.io/display-name: Fancy Chart Name Here # The chart's name you want displayed in the UI
.....

catalog.cattle.io/type: cluster-tool is used to list the chart in the cluster tools page and should be used only for the rancher cluster tools and NOT any partner charts.

Please test it from your side first on the Rancher and make sure no errors during deployment.
Again, I'm not testing the functionality of your chart, it's all on your side.
I just need to be able to deploy it from the Rancher side. Usually I'm testing on both; k3s and rke. If for some reason it's not possible to deploy on k3s from the Rancher side, just put limitation/requirements in you readme file for that.

alex-isv:
In addition, these are the description on asked annotation:

catalog.cattle.io/namespace: kube-system -- the default namespace where the chart to be installed is kube-system
catalog.cattle.io/os: linux              -- the chart can be installed into the cluster which nodes runs only Linux
catalog.cattle.io/scope: downstream      -- show the chart only on downstream clusters

They all valid.

eliaskoromilas:
> They all valid.

Thanks for the verification.

> Your chart version is still listing +rancher
> Please test it from your side first on the Rancher and make sure no errors during deployment.

I have already fixed (and tested) this, as I mentioned. I suspect, the problem on your end is that, since `2.5.1+rancher` > `2.5.1`, you have to get rid of the previous chart before installing the new one. Then you will be able to deploy it successfully.

> catalog.cattle.io/type: cluster-tool is used to list the chart in the cluster tools page and should be used only for the rancher cluster tools and NOT any partner charts.

I will remove this annotation if it's against the rules, or I could leave it as is If it's ignored anyway.

> I just need to be able to deploy it from the Rancher side. Usually I'm testing on both; k3s and rke. If for some reason it's not possible to deploy on k3s from the Rancher side, just put limitation/requirements in you readme file for that.

I understand your confusion, but really there is no limitation deploying our chart in a k3s cluster. I explained that the reason you can't see it in your **local** k3s cluster is that this cluster is considered `upstream`. The `catalog.cattle.io/scope: downstream` annotation enables our chart only on `downstream` (a.k.a **user**) clusters. There is really nothing that we could add as an explanation in the `app-readme`, since either the user can see the chart in the chart list (`downstream` clusters) or not (`upstream` clusters). This is explained in the link that I shared with you earlier.

alex-isv (changes requested):
@eliaskoromilas 

I was able to deploy in RKE, but I can see some hangs on DaemonSet - fpga-operator 

image

Can you clarify if any changes in config. is needed to pass it.

alex-isv:
Ignore my last update. It took a little longer to show it as active.

eliaskoromilas:
Thanks @alex-isv ! The chart can be deployed with the default parameters.

Note: You may see coral (daemonset) containers restart if the respective node doesn't have FPGA resources. It's normal.

alex-isv (dismissed):

alex-isv (changes requested):
@eliaskoromilas 
package.yaml should have 2 lines:

image

Please review -> https://github.com/alex-isv/partner-charts/tree/main-source#2-set-up-your-package-to-track-an-upstream-chart-skip-if-upgrading-existing-package

you just need to add packageVersion: 00

eliaskoromilas:
Thanks for pointing this out. It should be fine now.

alex-isv:
@eliaskoromilas 
Can you try to deploy and verify again from your side as It's hanging on my side after you made your last commit.

eliaskoromilas:
I verify that the new chart can be deployed successfully on Rancher v2.6.0.

```
---------------------------------------------------------------------
SUCCESS: helm upgrade --history-max=5 --install=true --namespace=kube-system --timeout=10m0s --values=/home/shell/helm/values-fpga-operator-2.5.201.yaml --version=2.5.201 --wait=true inaccel /home/shell/helm/fpga-operator-2.5.201.tgz
---------------------------------------------------------------------
```

eliaskoromilas:
rebase

alex-isv (changes requested):
Hi @eliaskoromilas 
Please do the following:
delete your last 'make charts' (your name) commit
rerun make charts and re-commit
rerun #make validate
make sure no errors at the end.
That should clean the old errors from validate.

eliaskoromilas:
@alex-isv I don't understand how recommiting the same changes will solve the validation issue.

After my previous rebase, `make validate` fails with the following error:

```
FATA[0056] Encountered error while trying to prepare package: Encountered error while preparing main chart: Encountered error while trying to pull u
pstream into charts: Unable to get chart archive: Get "https://helm.linkerd.io/stable/linkerd2-2.11.0.tgz": dial tcp 104.21.63.202:443: i/o timeout
```

Before that, `make validate` used to fail because of an issue with the `federatorai` chart. I rebased again today, but still the above `linkerd2` issue is there.

I tried without `linkerd2` chart and the validation finishes successfully.

alex-isv (changes requested):
@eliaskoromilas 
  catalog.cattle.io/type: cluster-tool should not be used in any partner charts
Please remove it and recommit.
Rerun make validate and make sure no errors.
Make sure you have 2 commits at the end.

alex-isv:
> > catalog.cattle.io/type: cluster-tool should not be used in any partner charts
> 
> Thanks for clarifying this requirement.
> 
> > Please remove it and recommit.
> > Rerun make validate and make sure no errors.
> > Make sure you have 2 commits at the end.
> 
> This is done.
> 
> @alex-isv Could we pull the Rancher-specific annotations and the `app-readme.md` to the upstream chart? We would like to automate the downstream chart update process. By doing so, no `generated-changes` with `patch` & `overlay` directories would be required.

@eliaskoromilas 
Yes, you can do it, but you still need to submit PR to update the URL in the package.yaml when a new chart is released, so we can pull it.
However, we won't be able to review any changes to the rancher specific things, including things we might request in the future because the changes aren’t being done in our repository and we won’t be able to see what changed. This will complicate the review process for rancher specific things, and any problems that happen would be your responsibility.

Also, can you verify if any other changes was done from your side? I'm on your current commit but I cannot deploy your chart. I tried 2 different RKE clusters and in both cases it's just hanging and not doing anything. Do you have any prereq. to be done before the installation or any specific parameters changes on the Rancher side? Can you lead me through the process. I know it worked before, but after some recent changes on your chart, I cannot deploy it anymore. And I don't have any errors generating, it's just hanging with the following:

image

alex-isv (approved):
@eliaskoromilas 
Without helm wait I can deploy now.
Looks good

PR #181, opened 2021-10-06
Portworx charts v2.8.0

dahuang-purestorage:
contains latest portworx charts v2.8.0. Any feedback is appreciated. thanks!


harsh-px:
@PennyScissors can we help with moving this forward?

dahuang-purestorage:
@alex-isv 
I have tried manually updating the folder to the format {VERSION}{PackageVersion}, but this would break when I run `make validate`. 

I think there are some edge case I'm running into with the `charts-build-script` or I'm missing some configuration setup. Please let me know what I can do to fix this.
Many thanks!

alex-isv:
@dahuang-purestorage 
Looks like now, Rancher is forcing to have a valid semver. So, in your case 2.8.1 is a valid one, so you should be good.

alex-isv:
@dahuang-purestorage 
Please clarify which parameters to use to install it from the local RKE cluster.

dahuang-purestorage:
@alex-isv 
this is the command i use for helm3 installation:
```
helm install --debug  --namespace kube-system --set etcdType=Built-in,environment="OnPrem",onpremStorage="Automatically scan disks",usedrivesAndPartitions=true --generate-name charts/portworx/portworx/2.8.1/
```

is this what you are looking for? Is there a way we can document this in partner-charts?
thanks!

alex-isv (approved):
Looks good

Verified on RKE

image

dahuang-purestorage:
@alex-isv friendly reminder, when might this be merged? Let me know if you need more info from my end. thanks!

PR #183, opened 2021-10-06
CloudCasa release v2.1.0

nitishdsharma:
**Updates**

PV backup to CloudCasa
SafeLock
EKS cluster auto-discovery
Application hooks
API keys
Multi-factor authentication


alex-isv (changes requested):
@nitishdsharma 

.orig and .rej under overlay should not be there

image

Also, provide steps and prereq. on installation

image

What's the cluster ID?

PR #192, opened 2021-10-08
Nutanix 2.4.1 rev

tuxtof:


alex-isv (changes requested):
Hi @tuxtof 
Overall looks good and I tested on rke.
We got some changes in rancher/main, so validate is showing as failed.

Please do the following:
delete your last 'make charts' (your name) commit
rerun make charts and re-commit
rerun #make validate
make sure no errors at the end.
That should clean the old errors from validate.

PR #193, opened 2021-10-11
citrix-adc-istio-ingress-gateway to version 1.11.0 

ajeetas:


ajeetas:
Hi @alex-isv,
I have incorporated your review comments.

Thanks,
Ajeeta

alex-isv (changes requested):
@ajeetas 
You need to have only 2 commits in your PR.
Make sure that 'make validate' has no errors from your side.
Also, can you verify if you can deploy it from your side because I tried from k3s and rke and it's hanging on both. Are there any additional settings to implement from my side in addition to Eula and PortNode?

ajeetas:
Hi @alex-isv,

Yes that `make validate` creating issue for `federatorai `.
Should I need created new PR with just 2 commit, as I have did in the very first two commits in this PR?
Thanks,
Ajeeta

PR #209, opened 2021-10-20
(Trilio) TrilioVault for Kubernetes Operator Release Update v2.5.3

bhagirathhapse:
This PR consists of: 

1. URL change in `package.yaml` for `2.5.2` release update
2. Annotations for partner, package-name and display-name in `Chart.yaml`
3. Newly generated `Chart.yaml.patch` file
4. Newly generated `assets/` and `charts/` directories.


alex-isv:
@bhagirathhapse 
The definition is wrong for the packageVersion.
Using version as per Rancher example is required and no exceptions here. 
-> https://github.com/rancher/partner-charts/tree/main-source#2-set-up-your-package-to-track-an-upstream-chart-skip-if-upgrading-existing-package

url: https://example.com/helm-repo/chart-0.1.2.tgz
packageVersion: 01

All charts are working this way except yours.
When I tested your tarball url with a packageVersion 01 I got k8s-triliovault-operator-2.5.201.tgz
Most was deployed from the Rancher, except k8s-triliovault-operator 

image

and it's erroring out with 

image

I don't see generated logs on my side.
Can you verify on your side what is related to that webhook-init and why is it hanging.
Either there are some settings as a  prereq. that I'm not aware of and you need to clarify that for me as it's listing 'Deployment does not have minimum availability. '
 or that component probably has a template for the chart version and breaks if it’s not the same as upstream. So, that is what you need to find from your side. Some values from your side are not right and when comparing version just failing because not available on your repo. So, if it worked for you before it doesn't mean it was correct. 
 I'm not sure if you can try 2.6.0 for your tarball and 00 for the packageVersion, and see if version will remain as 2.6.0 and won't complain when pulling from your repo.

bhagirathhapse:
Thank you @samuelattwood for the review. I will be creating new PR for TVK operator 2.5.3!

bhagirathhapse:
New PR #[229](https://github.com/rancher/partner-charts/pull/229) for reference.

PR #221, opened 2021-10-25
Shipa init clean

ravilach:
Hope third time is the charm, ha!


ravilach:
Chatting with Sam, will close this PR and update with the updated questions and readme. Tested on K3. 
-Ravi

PR #224, opened 2021-10-31
Add SoftIron RBD Chart

dabukalam:
Signed-off-by: Danny Abukalam <danny@softiron.com>


alex-isv (changes requested):
Hi @dabukalam 
Please review the workflow from -> https://github.com/rancher/partner-charts/tree/main-source#workflow
You need to have the same structure as in our example with asset, chart, index, etc. and you need to have 2 commits in your PR at the end with 1st commit 'changes' and 2nd commit 'make charts'
A good example should look similar to this -> 
image

dabukalam:
@alex-isv just updated this, let me know if that's what you're looking for

alex-isv (changes requested):
Hi @dabukalam 
Can you clarify steps on installation. Which setup is required in order to install it from Rancher?
Please provide detailed steps. Just remember that usually we just testing installation from the local instance and if any additional setup is needed you need to provide details on that.

dabukalam:
@alex-isv @samuelattwood I've just pushed the changes we discussed yesterday for your review - let me know if it works your end.

samuelattwood (approved):
Thanks Danny

Tested on rke2 1.22.4 and k3s 1.22.4. Looks good - Approved.

alex-isv (approved):

PR #234, opened 2021-11-03
Updating NeuVector chart version 1.8.5

selvamt94:


samuelattwood:
I have tested this chart on k3s (containerd), RKE2 (containerd), and RKE (Docker). When run on the containerd runtime, the controller and enforcer pods enter a crash loop due to `|ERRO|CTL|main.main: Failed to initialize - error=Unknown container runtime`. The scanner and manager pods appear to start normally. When run on RKE with the Docker runtime everything deploys as expected.

selvamt94:
Default container runtime is set to docker, so please disable docker and enable containerd in the values.yaml for containerd environment while deploying the chart. This option is not available in the questions.yaml. Please let me know if you still have issues. Thanks.
docker:
  enabled: false
  path: /var/run/docker.sock
containerd:
  enabled: true
  path: /var/run/containerd/containerd.sock

samuelattwood (approved):
Thanks, RKE2 and k3s are confirmed working with containerd runtime when using the k3s or containerd values.yaml options.

This may want to be more clearly exposed via the questions.yaml in the future, given that containerd is increasingly supplanting Docker as runtime.

alex-isv (approved):
approved

PR #254, opened 2021-11-22
Adding Digitalis vals-operator

digiserg:
Adds helm chart for [vals-operator](https://github.com/digitalis-io/vals-operator)


digiserg:
I'll create it again shortly as we have 0.2.0 released

PR #257, opened 2021-11-23
Vals operator 0.2.0

digiserg:
vals-operator syncs secrets from any secrets store supported by vals into Kubernetes. It works very similarly to secrets-manager and the code is actually based on it. Where they differ is that it not just supports HashiCorp Vault but many other secrets stores.

https://github.com/digitalis-io/vals-operator


alex-isv (changes requested):
Overall looks good.
In your chart.yaml change your catalog.cattle.io/display-name: vals-operator
to capital Vals-Operator
and make sure you have only 2 commits in your PR at the end.

alex-isv (changes requested):
Due to unexpected changes from upstream our validation was failing. Looks like your PR was created during that time. We fixed the issue but you need to revert back and sync with Rancher branches again. Make sure that git status is clean before commits. You need to have 2 commits at the end in your PR.

Thanks

samuelattwood:
It looks like you've created an altered git history in your fork. Please ensure that you are working off a clean and up-to-date fork and only include commits relevant to your chart.

digiserg:
> It looks like you've created an altered git history in your fork. Please ensure that you are working off a clean and up-to-date fork and only include commits relevant to your chart.

I messed up the rebase, sorry about that. It should be sorted out now.

samuelattwood:
You're close but just a couple fixes are necessary.

Please ensure you don't capture the `packages/vals-operator/charts` directory in your commit. That directory is removed  by `make clean` which should be run prior to your first commit, right after `make patch`.

To clarify what Alex said above, you likely want `catalog.cattle.io/display-name` set to something like `Vals Operator` or `Vals-Operator`, however `catalog.cattle.io/release-name` needs to be lowercase and without spaces. The `display-name` is what will appear in the marketplace UI beside the icon. It is up to you, but I would expect you would want something like this:

```
catalog.cattle.io/release-name: vals-operator
catalog.cattle.io/display-name: Vals-Operator
```

digiserg:
Thank you @samuelattwood ; I'll get better at this :)

alex-isv (changes requested):
Your index.yaml doesn't have the same values as your Chart.yaml
it's missing annotations.
For catalog.cattle.io/display-name: 
just put Vals-Operator (without any further description) It should be a short chart name which will be displayed on the Marketplace.
Also, please have only 2 commits per PR.

PR #262, opened 2021-11-23
Speedscale Operator 0.9.126

swtch1:
This is the initial request to add the Speedscale operator chart.


samuelattwood:
This looks very close but there are a couple things of note.

Please make sure you don't capture packages/speedscale-operator/charts in your commit. This directory is transient and used for `make patch` but should get cleaned up by `make clean` prior to your first commit.

It looks like you've included the `app-readme.md`, `questions.yaml`, and `catalog.cattle.io` annotations in your upstream chart. From a functional perspective this will work fine, however we typically advise tracking these in this repo via the patch/overlay mechanism, rather than adding them directly to your upstream. This is ultimately a decision on your end, but that is our standard guidance.

swtch1:
Thanks Sam.  I've cleaned up the packages dir.  

We do not use helm anywhere else and created this chart for the Rancher integration so I think we'll leave those artifacts in for now, but that's a good call out.

Let me know if I need to make any further changes.

swtch1:
Sam, I'm assuming this will get pushed when you go to create the next release but I wanted to reach out early in case I need to make any further changes.

samuelattwood:
Hey Josh,
Sorry for the delay. Your file layout looks good now. Typically we prefer to test install the chart to ensure it deploys, but this doesn't seem to be possible here due to the API key requisite. Just so we have some degree of confirmation, if you could do me a favor and provide a screenshot of the successful helm install console from within the Rancher UI, I would appreciate it. We should be clear to merge otherwise. Thanks.

swtch1:
chart

swtch1:
install

swtch1:
operator-deploy

samuelattwood (approved):
Approved.

PR #265, opened 2021-11-26
Fix InAccel FPGA Operator

eliaskoromilas:
We encountered a problem with the last InAccel FPGA Operator downstream release. The issue is related to the way [rancher/charts-build-scripts](https://github.com/rancher/charts-build-scripts) handles chart dependencies and the produced downstream chart.

More specifically `make prepare` replaces all `dependency.repository` URLs with local paths but also removes the `dependency.version` fields. For example in our case:

```patch
<   repository: https://kubernetes-sigs.github.io/node-feature-discovery/charts
<   version: 0.9.0
---
>   repository: file://./charts/node-feature-discovery
```

But it seems that Helm ignores dependencies without a version field and instead just deploys everything under `charts` directory as a subchart (and not as a dependency), which means two things:

1. `dependency.alias` field is not taken into account, which is problematic if configuration is passed through `values.yaml`
2. `dependency.condition` is also ignored, which makes it impossible to disable the dependency (if needed, in a use-case)

As a workaround in this fix we restore the `dependency.version` field through the `Chart.yaml.patch`, but I believe that this should eventually be fixed in the `charts-build-scripts`. Do you want me to open a related issue there, too?

To replicate the issue run the following command and you will notice that every `fpga-discovery` related template appears although we disabled the dependency.

```
helm template --set fpga-discovery.enabled=false https://github.com/rancher/partner-charts/blob/main-source/assets/fpga-operator/fpga-operator-2.6.101.tgz?raw=true
```

Then run the same command but this time using the new chart package that just has the `dependency.version` field restored:

```
helm template --set fpga-discovery.enabled=false https://github.com/eliaskoromilas/partner-charts/blob/fpga-operator/assets/fpga-operator/fpga-operator-2.6.102.tgz?raw=true
```

Now the condition works but also the `fpga-discovery` configuration is applied by Helm.


alex-isv (approved):
Deployment verified on RKE

image

samuelattwood:
@eliaskoromilas Thanks for the detailed report. I am able to replicate the issue and agree that seems like that should be fixed in the build scripts themselves. I see you opened https://github.com/rancher/charts-build-scripts/issues/68 so hopefully we will have this addressed soon.

Fixing it in the patch file is a good workaround for the time being.

PR #270, opened 2021-11-30
Add KubeMQ charts

kubemq:


alex-isv (changes requested):
You are missing assets tarball and the index file.
Please review the workflow from > https://github.com/rancher/partner-charts#7-test-your-changes
You need to test it on your side first

PR #275, opened 2021-12-02
Update HAProxy to 1.17.10

andrewheberle:
This updates the HAProxy ingress controller chart to 1.17.10


alex-isv (approved):
verified on k3s and rke
image

PR #287, opened 2021-12-13
Initial TriggerMesh chart

mattray:
This provides the TriggerMesh Helm chart, I'll submit a separate PR to remove the `aws-event-sources` after this is merged.

I'm not sure of the formatting for the questions.yaml with regards to the hashes and arrays. The `values.yaml` that was the source looked like this:

```yaml
# Default values

nameOverride: ""
fullnameOverride: ""

replicaCount: 1

image:
  registry: "gcr.io/triggermesh"
  pullPolicy: "IfNotPresent"
  tag: ""

imagePullSecrets: []

rbac:
  create: true

serviceAccount:
  create: true
  annotations: {}
  name: ""

podAnnotations: {}

podSecurityContext: {}

securityContext:
  allowPrivilegeEscalation: false

resources: {}

nodeSelector: {}

tolerations: []

affinity: {}

webhook:
  podAnnotations:
    sidecar.istio.io/inject: 'false'

  podSecurityContext: {}

  securityContext: {}

  resources: {}

  nodeSelector: {}

  tolerations: []

  affinity: {}
```

Signed-off-by: Matt Ray <github@mattray.dev>


samuelattwood:
Hi @mattray. It looks like you only captured the files from the `make patch` output, and not the assets, charts, and index.yaml files generated by `make charts` so please add that additional commit. If you have any confusion after reviewing the [README](https://github.com/rancher/partner-charts/blob/main-source/README.md) please let me know.

You are also missing an end quote in your `Chart.yaml.patch` around your `kubeVersion`

Regarding the `questions.yaml`, currently dicts and lists are not supported in this file, so as-is if you need  user-configured variables of those types, the 'Edit YAML' feature will need to be used. You also need to include a `label` for each variable, which is the name the variable appears as within the UI. I will also mention that if a variable is present already with a default value in the chart's `values.yaml`, Rancher will present that as the default value, so setting a `default` in the `questions.yaml` is not strictly required.

mattray:
@samuelattwood I started fresh with https://github.com/rancher/partner-charts/pull/316 since this was pretty stale.

PR #288, opened 2021-12-13
Committing StackState Charts

anthonyStacked:


samuelattwood:
Hi @anthonyStacked. It looks like you only committed the files following `make patch` and not the charts, assets, and index.yaml generated by `make charts` so please add that second commit. Review the process in the [README](https://github.com/rancher/partner-charts/blob/main-source/README.md) and let me know if you have any confusion.

A couple other things of note:
- The `kubeVersion` variable should be added to the Chart.yaml
- Your icon URL currently returns a 404

This is optional, but we recommend using the `questions.yaml` mechanism for chart configuration, especially for required values. Since your chart only mentions 3 required values, this would be very easy to include. Yours could look something like this

```
questions:
- variable: stackstate.apiKey
  label: "API Key"
  required: true
  type: string
- variable: stackstate.cluster.name
  label: "Cluster Name"
  required: true
  type: string
- variable: stackstate.url
  label: "URL"
  required: true
  type: string
```

samuelattwood:
@anthonyStacked I am going to close this out as it has been sitting for some time without update. We would welcome a fresh PR from StackState in the future. Let me know if there is anything I can help with. Thanks

PR #296, opened 2021-12-24
(Trilio) TrilioVault for Kubernetes Operator Release Update v2.6.4

bhagirathhapse:
This PR consists of:
1. URL change in package.yaml for 2.6.4 release update
2. Annotations for partner, package-name, and display-name in Chart.yaml
3. Newly generated Chart.yaml.patch file
4. Newly generated assets/ and charts/ directories.


samuelattwood:
Hey @bhagirathhapse, thanks for the update. One small thing before we merge, it is necessary for the `kubeVersion` range to have spaces around the `-` otherwise it is interpreted as one version, so `1.18 - 1.22` rather than `1.18-1.22`. I tested your chart with this fix and it deployed without issue so once that's fixed we should be all set.

bhagirathhapse:
Thank you @samuelattwood for the review comment.
I have made the changes and pushed the code using `git push --force` to make sure it replaces the files. I had to revert the `index.yaml` since the `make chart` command does not run with any changes present in the local repo. I think this should be good to merge now.
Let me know if you have any suggestions.

samuelattwood (approved):
Confirmed that asset tarball, chart, and index.yaml match following force push.

Test deployment ran cleanly on k3s 1.22.4. Approved.

PR #297, opened 2021-12-28
Neuvector: updating NeuVector chart version to 1.8.8

selvamt94:


samuelattwood (approved):
Layout looks good. Tested on k3s 1.22.4 (Selected containerd runtime with socket path as /var/run/k3s/containerd/containerd.sock). Approved.

PR #300, opened 2022-01-05
Nutanix csi storage v2.5.0

tuxtof:

PR #302, opened 2022-01-05
Nutanix csi snapshot 1.0.0

tuxtof:


samuelattwood:
Hey @tuxtof 
It looks like this chart, and upon review also #300, leverage `CustomResourceDefinition` from `apiextensions.k8s.io/v1beta1`. As of Kubernetes 1.16 this resource has been available from `apiextensions.k8s.io/v1` and as of [Kubernetes 1.22](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22), the v1beta1 version of the resource is no longer served. This should be a very easy fix on your end, but it does mean that currently the charts will not install properly on Kubernetes >= 1.22. You may want to fix this API version and/or set your `kubeVersion` to a range i.e. `kubeVersion: 1.17 - 1.21`

tuxtof:
Hey @samuelattwood , where did you see i use `apiextensions.k8s.io/v1beta1` ???
All my crd are served from `apiextensions.k8s.io/v1` and the chart works perfectly with k8s >= 1.22

thanks

samuelattwood (approved):
You are correct. It looks like the CSI Storage chart version 2.4.1 may have used the v1beta1, I must have confused charts when I ran this test. Apologies.

Test deployment ran cleanly on RKE2 1.22.4 and k3s 1.22.4. Approved.

tuxtof:
yes i confirm it was the case with the previous version

thanks @samuelattwood

PR #304, opened 2022-01-07
Vals operator 0.3.0

digiserg:
https://github.com/digitalis-io/vals-operator/releases/tag/v0.4.0


samuelattwood:
Hey @digiserg 
It looks like you captured the temporary files `Chart.yaml.orig` and `Chart.yaml.rej`. These can get generated when `make prepare` is run and the patch files aren't able to be applied due to changes in the Chart.yaml. It also looks like you've removed your app-readme.md, questions.yaml, and Chart.yaml.patch which probably was not intended.

If you're having an issue with `make prepare` creating the `.orig` and `.rej` files, you can remove `packages/vals-operator/generated-changes/patch/Chart.yaml.patch`. `make patch` will make a new one for you anyway. Let me know if there is anything I can clarify.

digiserg:
Thank you @samuelattwood , I think I completely misunderstood how the package is updated. I was expecting all the original settings to be preserved such as questions or app-readme and didn't even check. I hope it's correct now.

samuelattwood:
Hey @digiserg. You fixed the Chart.yaml issues, but it still looks like your `app-readme.md` and `questions.yaml` are getting removed.

samuelattwood:
Hey @digiserg You're very close but those two files still are removed from the overlay directory. To explain this process a bit, in `packages/vals-operator/generated-changes` there are typically two directories, `overlay` and `patch`. 

Files in `overlay` will be added in to the chart directory and gzip, this is where the `questions.yaml` and `app-readme.md` are placed when you run `make patch`. Files in `patch` contains patch files that are able to edit text, such as adding the Rancher annotations, in pre-existing chart files.

When `make charts` is run, your upstream chart url is fetched, files in `overlay` are dropped in, edits from `patch` are applied, then this is bundled into a gzip as well as placed in the charts directory. Hopefully this helps.

digiserg:
I believe I found the problem, it was being excluded because my global `gitignore`. I hope I finally managed to get it right.

samuelattwood (approved):
Hey @digiserg, I believe I made a mistake in my last comment. You had fixed those files being removed, it was just that they also were not being modified so they did not appear in the PR. It looks fine now as well, but apologies for the confusion.

Ran test deployment on k3s 1.22.5 and RKE 1.21.6. Approved.

digiserg:
Excellent, thank you @samuelattwood 

The only thing I just noticed is we're partners but don't see the partner icon next to the app logo. What am I doing wrong?

PR #315, opened 2022-01-14
Add K10 chart version 4.5.6

A-kanksh-a:
This PR is to add K10 chart version 4.5.6 to rancher Marketplace.


samuelattwood:
@A-kanksh-a Thank you for the submission. Your PR looks good and a test deployment on k3s 1.22.4 ran without issue.

For chart versioning we require a `packageVersion` variable in the `package.yaml`which allows for independent versioning of the chart and the application. So if you, for example, needed to tweak something about the chart but wanted to remain on K10 4.5.6 you can bump the `packageVersion` and leave the `appVersion` the same Typically it is set initially to `00`.

This calculated version sets the `.Chart.Version` variable however, and it looks like in your `templates/_k10_serviceimage.tpl` you use `.Chart.Version` for the image tag, which this would currently break. It would be looking for an image tagged, for example, `4.5.600` rather than `4.5.6`. When I replaced `.Chart.Version` with `.Chart.AppVersion` for your image tag, the issue was rectified.

Everything else looks good so as soon as we can sort this, we will get the chart merged. Please let me know if there is anything I can clarify. Thanks.

A-kanksh-a:
thanks for the reviews @samuelattwood

Is it a required field to have `packageVersion` field in `package.yaml`? In our scenario, If we have to modify the chart, we generally release a completely new version itself. So every time we will have a new upstream `url` in case of any modification. 

As you mentioned, If we add `packageVersion`, it will recalculate the `version` but in our case, we are using the `version` as an image tag. So this would break the chart deployment.

A-kanksh-a:
@samuelattwood Could you please revert back with your suggestions here?

samuelattwood:
@A-kanksh-a Unfortunately we do require the `packageVersion` variable in an effort to follow semantic versioning and allow iteration on specifically the chart version when necessary.

The fix on your end should be fairly simple since it appears you are already using `appVersion` and `version` as tracking the same value. If you simply switch to using `.Chart.appVersion` for your image tag instead of `.Chart.Version` this would resolve the issue. Per [Helm documentation](https://helm.sh/docs/topics/charts/#charts-and-versioning), `appVersion` is intended to track the version of the underlying application, whereas `version` is meant to track the version of the chart itself.

A-kanksh-a:
Hey @samuelattwood, here is the new PR for the `K10` chart [https://github.com/rancher/partner-charts/pull/340](https://github.com/rancher/partner-charts/pull/340) which uses `packageVersion` in `package.yaml`

Also, I am closing this PR as this k10 version (4.5.6) doesn't use `.Chart.appVersion` as image tag. We have updated the chart to use `AppVersion` as image tag in the latest release.

A-kanksh-a:
New PR [https://github.com/rancher/partner-charts/pull/340](https://github.com/rancher/partner-charts/pull/340)

PR #320, opened 2022-01-21
Add ondat 0.5.2

DavidMarchant:
Add ondat-operator chart to new repo

**Previously known as StorageOS**
We were present in the old repo here: https://github.com/rancher/charts/tree/master/charts/storageos-operator. You can read more about our name change here: https://www.ondat.io/blog/welcome-to-ondat
Our new charts repo lives here: https://github.com/ondat/charts

Please let me know if you need any more confirmation that this is the same chart that was previously certified as partner! Or indeed of any changes to this PR


samuelattwood (approved):
Looks good. Ran test deployment on rke 1.21.6 and k3s 1.22.5 successfully using external etcd. Approved.

PR #324, opened 2022-01-24
Kongmesh branch

CAcquaviva:


samuelattwood:
Hey @CAcquaviva,
Your PR looks good and deploys just fine, but you may want to just drop your `questions.yaml` file, the `labels:` and `categories:` variables no longer serve a function here, and you don't appear to be otherwise using the `questions.yaml` for configuration of your chart, so you can simply drop that file. Having the empty questions list is causing an odd rendering of UI elements (which I am looking in to).

CAcquaviva:
Thanks for the instructions. Submitting a new PR now.

PR #327, opened 2022-01-31
AWS event source has been replaced by the TriggerMesh chart

mattray:
The AWS event source has been deprecated in favor of the TriggerMesh chart which includes it. We no longer break this package out separately.


samuelattwood:
Currently we prefer to not outright remove charts from the repository, but we have a mechanism to hide charts from the UI with an annotation. This chart has been marked as hidden in #412 and will no longer appear in Rancher. Let me know if you have any concerns. Thanks

PR #351, opened 2022-02-24
Add ondat 0.5.4

DavidMarchant:
Version v2.6.0 of Ondat


samuelattwood (approved):
Looks good. Tested on k3s 1.22.5 and rke 1.21.6. Approved.

PR #318, opened 2022-01-21
Add universal-crossplane 1.6.1001

turkenh:
This PR adds Universal Crossplane [v1.6.1-up.1](https://github.com/upbound/universal-crossplane/releases/tag/v1.6.1-up.1).

PR #101, opened 2021-06-25
(Upbound) Update Universal Crossplane to v1.2.3-up.2

turkenh:
This PR updates Upbound's Universal Crossplane chart to its latest version. See the release notes for more details:

https://github.com/upbound/universal-crossplane/releases/tag/v1.2.3-up.2

PR #353, opened 2022-02-27
universal-crossplane: add v1.4.4001

hasheddan:
Adds universal-crossplane v1.4.4001.


samuelattwood:
Hi @hasheddan It looks like you're doing some unexpected things with altering versioning so I just wanted to clarify if this is intentional.

It looks like via the `Chart.yaml.patch` you are modifying both `appVersion` and `version`. This is not typically expected and by manually setting `version` to `1.4.4001` the resultant version from Rancher will be `1.4.400101`. The chart version is calculated as `int(patchVersion*100 + packageVersion)` where `packageVersion` comes from the `package.yaml`.

`appVersion` is not altered in any way by Rancher and generally is expected to be unchanged as it is often used in the chart itself for things like image tags.

As an aside, you are likely to cause a merge conflict with yourself by submitting multiple PRs simultaneously based on the same initial commit, as they will be editing the same location in the `index.yaml`

Please let me know if there is anything I can help with or clarify further. Thanks

hasheddan:
@samuelattwood thanks for following up! Looking back at previous updates, specifically [the most recent one](https://github.com/rancher/partner-charts/pull/318/commits/af64d813d90c6ec755c37249bf020646a4754aef), it appears this was the strategy followed. Could you point out to me which parts you would expect to look different in my PR?

> As an aside, you are likely to cause a merge conflict with yourself by submitting multiple PRs simultaneously based on the same initial commit, as they will be editing the same location in the index.yaml

Yep, just opening them all for linking to our relevant release issues. Happy to rebase once their is alignment on their content.

samuelattwood:
@hasheddan I ought to have mentioned it in the previous PR as well, but it won't hurt anything. Largely it is just that your `Chart.yaml.patch` includes alterations for `version` and `appVersion`. This means that these variables were (probably unnecessarily) edited in the Chart.yaml prior to running `make patch`.

The items that are typically expected to be altered by this mechanism are the `catalog.cattle.io` annotations and, if not present in your upstream, `kubeVersion` and `icon`.

Unless you have a reason to alter any other variables, what you can do is just remove `packages/universal-crossplane/generated-changes/patch/Chart.yaml.patch` and have a new one be generated with just the necessary information.

Starting with where your branch is currently, [here is a brief recording](https://asciinema.org/a/bhL1SXwoVEhZsLF1JHGmCs28c) of how you can restore a state we would typically expect. These are not steps you should need to repeat in the future, it is just to clean up the current state.

Please let me know if there is anything I can help with or clarify further.

hasheddan:
@samuelattwood thanks for making and sharing a recording! I think we may be missing some shared context here, which may be clarified by taking a look at the initial PR that moved to the strategy being used here (https://github.com/rancher/partner-charts/pull/101). I believe the comments made in conversation in that PR still hold true. Could you let me know if that is no longer the case? Thanks!

samuelattwood:
@hasheddan Thanks for the added context. You guys break down the root cause well in your own issue [here](https://github.com/upbound/universal-crossplane/issues/119), which clarifies this. Rancher's default hiding of pre-release versions, as defined by SemVer2, is still true so feel free to continue using `00` as a sort of delimiter to bypass that issue. 

I will proceed with standard review. Thanks :slightly_smiling_face:

samuelattwood (approved):
Tested on k3s 1.22.6 and rke2 1.21.9. Approved.

PR #362, opened 2022-03-07
Add openziti package to partner charts

qrkourier:
Thanks for adding us! Let me know if anything's not perfectly according to spec in https://github.com/rancher/partner-charts.


samuelattwood:
@qrkourier It looks like you may have accidentally left in a testing IP for your icon in the `Chart.yaml`. Please also run `make clean` prior to your first commit, the `packages/ziti-host/charts` directory is not intended to get committed, and `make clean` will remove it.

Other than that you should be good to go. I have tested on k3s 1.22.6 and RKE2 1.22.5 successfully using an externally hosted controller and edge router using the [one-liner setup](https://openziti.github.io/ziti/quickstarts/network/local-no-docker.html). Let me know if I can clarify anything. Thanks

qrkourier (commented):
@samuelattwood Thank you. Now I see step 7 is only for modifications. Hopefully I got it right this time! Would you be willing to give us a quick feedback about your experience with the OpenZiti quickstart or the chart in [GitHub](https://github.com/openziti/ziti/issues) or [Discourse](https://openziti.discourse.group/)?

samuelattwood (approved):
Looks good, thanks for the update. Approved.

Issue #235, opened 2021-11-04
No access to any values with dash in the name


moabu:
🐛 🐛 🐛  
#### Steps to reproduce 

1. Create a `questions.yaml` with the following

```yaml
# MyExampleApp
- variable: global.myapp-api.enabled
  default: true
  type: boolean
  group: "Optional Services"
  required: true
  label: Enable MyAppAPI
  description: "What a cool app!"
- variable: myapi-app.replicas
  default: 1
  required: false
  type: int
  group: "Replicas"
  label: MyAppAPI Replicas
  description: "Service replica number."
  show_if: "global.myapp-api.enabled=true"
```
  
2. Follow [instructions](https://github.com/rancher/partner-charts/tree/main-source) to generate and test the chart . Add the repository to the UI.

3. Inside the UI, head to the Charts section and click on the chart added. Choose the option enabling the example app. 

4. The Replicas group does not appear. 

This happens with all dashed keys.

#### Current behaviour (bug)

It seems like any key that contains a dash `-` in the `questions.yaml` is not accessed, and hence does not resolve correctly in the UI. 

```yaml
# MyExampleApp
- variable: global.myapp-api.enabled
  default: true
  type: boolean
  group: "Optional Services"
  required: true
  label: Enable MyAppAPI
  description: "What a cool app!"
- variable: myapi-app.replicas
  default: 1
  required: false
  type: int
  group: "Replicas"
  label: MyAppAPI Replicas
  description: "Service replica number."
  show_if: "global.myapp-api.enabled=true"
```

#### Expected behaviour (correct)

I expect the UI `show_if`  and `show_subquestion_if` to respond to dashed keys.

#### Relevant Rancher logs

```
HEAD is now at 8a66d84 test:re
2021-11-04 08:36:45.202767 I | mvcc: store.index: compact 41063
2021-11-04 08:36:45.240412 I | mvcc: finished scheduled compaction at 41063 (took 36.708926ms)
2021/11/04 08:36:45 [TRACE] REST LIST api//v1//componentstatuses
2021/11/04 08:37:00 [TRACE] REST LIST api//v1//componentstatuses
2021/11/04 08:37:15 [TRACE] REST LIST api//v1//componentstatuses
HEAD is now at cc5a839 Merge pull request #228 from alex-isv/main
HEAD is now at cc5a839 Merge pull request #228 from alex-isv/main
HEAD is now at 1a6477e Bump harvester csi driver 0.1.4
[...]

PR #370, opened 2022-03-16
feat(gluu): update Gluu helm partner chart

mo-auto:
# Updating Gluu flex Partner Chart

This is an automated PR triggered by a release in [Gluu](https://github.com/GluuFederation/flex) charts.

## Description

This PR updates the Gluu official chart to rancher ui partner charts.  We reported an [issue](https://github.com/rancher/partner-charts/issues/235) regarding using dashed keys inside `questions.yaml`. 

## Type of change

- [x] Update Chart (non-breaking change which adds functionality)

## How Has This Been Tested?

- [x] Followed instructions regarding testing, using a docker based rancher ui setup. 

**Test Configuration**:
OS Ubuntu 20 LTS with docker 20.10. 7 and rancher 2.6.

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings during validation
- [x] I have checked my code and corrected any misspellings


moabu:
@samuelattwood csn we get this Par checked please.

samuelattwood:
@moabu Yes, it will be reviewed today.

moabu:
Thank you !

samuelattwood (approved):
Tested on RKE2 1.22.5. Approved.

@moabu Two things I'll note:
- You don't need to alter `packageVersion` if the version of the chart is already changing
- I was a bit confused by the several `questions.yaml` values that say `Enable this!` in their description but default to `false`

moabu:
- ok perfect we won't changed the packaged version then. We had thought it's something internally used by the UI
- it only shows Enable this only if the associated service is enabled. Because the user has by default the option to disable the endpoints associated with the ingress but we recommend enabling them publicly.

samuelattwood:
@moabu Ah, alright I understand. Typically you should not even need to provide a `default` value in the `questions.yaml` if you are setting a default in your `values.yaml`, since that will override anyway. This partially answers your question, in that the `questions.yaml` default will never override the `values.yaml` default.

What you may be able to do, is resolve this using Helm mechanics. As an example, if we looks at `nginx-ingress.ingress.fido2ConfigEnabled` you could likely do something like this in your `charts/nginx-ingress/templates/ingress.yaml`
```
{{ if or (and (eq .Values.global.distribution "default") (not .Values.global.istio.ingress) (.Values.global.fido2.enabled)) (.Values.ingress.fido2ConfigEnabled) -}}
```
I haven't explicitly tested this but in theory it should do what I believe you're looking for, albeit a bit verbosely.

Would you like me to proceed with merging this chart as-is?

PR #375, opened 2022-03-24
Nutanix CSI storage 2.5.2

tuxtof:


samuelattwood:
@tuxtof Everything looks good and it deploys fine, but I did note that your `appVersion` is still set to `2.5.1`. Is that intended?

tuxtof:
@samuelattwood yes it is expected driver version don't change just the chart

samuelattwood (approved):
Sounds good, just wanted to make sure.
Tested on RKE2 1.22.5. Approved.

tuxtof:
Thanks for the remark

PR #381, opened 2022-04-01
(Trilio) TrilioVault for Kubernetes Operator Release Update v2.7.2 with improved questions.yaml implementation

bhagirathhapse:
This PR consists of:

1. URL change in `package.yaml` for 2.7.2 release update.
2. The updated `questions.yaml` file to felicitate users while adding the TVK instance name input parameter to `values.yaml` used by the TrilioVault helm chart
3. Annotations for partner, package-name, and display-name in `Chart.yaml`
4. Newly generated `Chart.yaml.patch` file
5. Newly generated `assets/` and `charts/` directories.

TVK Operator v2.7.2 also installs the TVK Manager of the same version by default on the same namespace and configures the TVK UI for users to start discovering their K8s cluster objects and start creating backups.


samuelattwood (approved):
Tested on k3s 1.22.2 and RKE 1.22.7. Approved.

Investigating issue when deploying on RKE2 with default Nginx ingress. Does not occur with k3s traefik or with RKE Nginx ingress.

PR #385, opened 2022-04-09
Update Kasten K10 Chart version to 4.5.13

infraq:
Update K10 Chart to 4.5.13


samuelattwood (approved):
Tested on RKE 1.22.7 and RKE2 1.22.7+rke2r2 with Rancher 2.6.4. Approved.

PR #387, opened 2022-04-13
Update portworx chart version to 2.9.1

dahuang-purestorage:


dahuang-purestorage:
@alex-isv review welcome.

samuelattwood:
Hey @dahuang-purestorage 
I have been trying to deploy this chart on a few different cluster configurations but I'm running into some snags. If I test on an EKS cluster, it tries to pull the image `gcr.io/google_containers/kube-scheduler-amd64:v1.21.4` for stork which doesn't appear to exist.

samuelattwood:
Hey @dahuang-purestorage
I have been trying to deploy this chart on a few different cluster configurations but I'm running into some snags. If I test on an EKS cluster, it tries to pull the image gcr.io/google_containers/kube-scheduler-amd64:v1.21.4 for stork which doesn't appear to exist.

When I deploy to an RKE or RKE2 cluster on EC2 instances, with additional EBS blocks, whether I specify the block device path or let it autodetect I seem to get something like this
```
level=error msg="Authentication error: Authentication error: NoCredentialProviders: no valid providers in chain. Deprecated.\n\tFor verbose messaging see aws.Config.CredentialsChainVerboseErrors" file="boot.go:142" component=porx/px/boot fields.func=InitAndBoot func= package=boot
level=error msg="Could not init boot manager" file="terminate.go:86" component=porx/pkg/util/logs error="Authentication error: Authentication error: NoCredentialProviders: no valid providers in chain. Deprecated.\n\tFor verbose messaging see aws.Config.CredentialsChainVerboseErrors"
PXPROCS[INFO]: px daemon exited with code: 1
INFO exited: pxdaemon (exit status 1; not expected)
```
The portworx daemon seems to install and run just fine, the underlying nodes are typically Ubuntu. Let me know if there is something obvious I'm overlooking or if there are different logs I should be checking.

dahuang-purestorage:
@samuelattwood 
thanks for raising the two issues. Quick question, what kubeversion are you running?

samuelattwood:
@dahuang-purestorage 
I just dug a bit further into this and realized the secgroup for my RKE cluster was too restrictive. Opening up the relevant port range resolved that issue, so apologies, that was my fault.

For EKS I was running `v1.21.5-eks-9017834` but I also tested upgrading to `v1.22.6-eks-7d68063` and had the same issue.

I believe this problem is due to a block in your `templates/_helpers.tpl` file. EKS doesn't correctly follow SemVer2 (Neither does GKE) as they use a `-` rather than a `+` to mark their builds. Most likely the semver comparison is getting tripped up because the EKS version is considered a pre-release due to that hyphen, and these images don't exist at `gcr.io/google_containers` they're at `k8s.gcr.io`.

In that same vein, you likely want to update your `kubeVersion` to be `kubeVersion: '>=1.16.0-0'`. Without that `-0` the chart will be considered incompatible with EKS and GKE versions due to what I describe above with the versioning. This is a [known issue](https://github.com/helm/helm/issues/10375). 

This is the relevant block.

```yaml
{{- define "px.getk8sImages" -}}
{{- $version := .Capabilities.KubeVersion.GitVersion -}}
{{- if (.Values.customRegistryURL) -}}
    {{- if (eq "/" (.Values.customRegistryURL | regexFind "/")) -}}
        {{ trim .Values.customRegistryURL }}
    {{- else -}}
        {{- if or (or (and (semverCompare ">= 1.16.14" $version ) (semverCompare "<=1.17.0"  $version)) (and (semverCompare ">= 1.17.10" $version) (semverCompare "<=1.18.0" $version ))) (semverCompare ">=1.18.7" $version) -}}
           {{cat (trim .Values.customRegistryURL) "/k8s.gcr.io" | replace " " ""}}
        {{- else -}}
           {{cat (trim .Values.customRegistryURL) "/gcr.io/google_containers" | replace " " ""}}
        {{- end -}}
    {{- end -}}
{{- else -}}
     {{- if or (or (and (semverCompare ">= 1.16.14" $version ) (semverCompare "<=1.17.0"  $version)) (and (semverCompare ">= 1.17.10" $version) (semverCompare "<=1.18.0" $version ))) (semverCompare ">=1.18.7" $version) -}}
        {{ "k8s.gcr.io" }}
     {{- else -}}
        {{ "gcr.io/google_containers" }}
    {{- end -}}
{{- end -}}
{{- end -}}
```

samuelattwood (approved):
@dahuang-purestorage I will proceed with merging this as these are pre-existing issues and are specific to EKS/GKE versioning. I'd definitely recommend trying to address them in a future release. Let me know if there is anything I can clarify or if I've overlooked anything. Thanks

Tested on RKE 1.22.9 with Rancher 2.6.4. Approved.

PR #396, opened 2022-04-15
HPE CSI Driver for Kubernetes Helm chart v2.1.1-0

datamattsson:


samuelattwood:
Hi @datamattsson,
I ran into two issues deploying the chart. Currently your new version is set to `2.1.1-0`. As a result of the `-0` this marks the chart as a pre-release version following [SemVer2](https://semver.org/#spec-item-9). By default, pre-release charts are hidden in Rancher.

The second thing is that your `questions.yaml` is setting an apparently now-unused variable of `cspClientTimeout` and due to your values schema verification this actually caused the install to fail.

Once I fixed those two small things, the chart installed without issue. Let me know if there is anything I can clarify. Thanks

datamattsson:
> Hi @datamattsson, I ran into two issues deploying the chart. Currently your new version is set to `2.1.1-0`. As a result of the `-0` this marks the chart as a pre-release version following [SemVer2](https://semver.org/#spec-item-9). By default, pre-release charts are hidden in Rancher.

Thanks for bringing this up. I'll take it up with our engineering, this explains some weird behavior I've seen elsewhere.

> The second thing is that your `questions.yaml` is setting an apparently now-unused variable of `cspClientTimeout` and due to your values schema verification this actually caused the install to fail.

Noted, I'll update accordingly.

> Once I fixed those two small things, the chart installed without issue. Let me know if there is anything I can clarify. Thanks

Would you need a new PR or may I add an additional commit to this PR?

samuelattwood:
@datamattsson Either is fine, if you'd like to continue in this PR feel free.

datamattsson:
I'll submit a new one.

PR #397, opened 2022-04-19
add DeployHub to Marketplace

sbtaylor15:
Adding DeployHub to the Marketplace as a new partner.


samuelattwood:
Hey @sbtaylor15
Thanks for the submission. I have a couple of questions/notes:

You may want to alter your `kubeVersion` to include `-0` at the end of each version. A couple of major platforms like GKE and and EKS don't completely comply with SemVer2 and use `-` as build delimiter instead of `+`, effectively marking their builds as pre-release. The result is Helm will consider the chart incompatible and it will be hidden in Rancher. A simple change to`kubeVersion: '1.19.0-0 - 1.23.0-0'` will fix this.

General note but it seems like your chart typically requires a loadbalancer resource in order to deploy. It looks like it will use NodePort if you specifically select `alb` for ingress but there are uses for NodePort and ClusterIP other than just an AWS Application Load Balancer, so that might be a useful option. 

Your nginx pod seems to fail to start if the namespace the chart is deployed to is anything other than `deployhub`. Your nginx image's entrypoint does do a find/replace in `/etc/nginx/nginx.conf` of `default.svc.cluster.local` based on the `POD_NAMESPACE` environment variable, but `/etc/nginx/nginx-ssl.conf` has some of the upstreams defaulting to `deployhub.svc.cluster.local` which then do not get replaced.

I have the chart deployed successfully with all pods running, but when I connect to the loadbalancer address I get an Apache Tomcat server error. It is connecting to the path `dmadminweb/Home` 
```
Type Exception Report

Description The server encountered an unexpected condition that prevented it from fulfilling the request.

Exception

java.lang.NullPointerException
	dmadmin.DMSession.getInstance(DMSession.java:472)
	dmadmin.HttpServletBase.internalHandleRequest(HttpServletBase.java:59)
	dmadmin.HttpServletBase.doGet(HttpServletBase.java:44)
	javax.servlet.http.HttpServlet.service(HttpServlet.java:635)
	javax.servlet.http.HttpServlet.service(HttpServlet.java:742)
```

sbtaylor15:
@samuelattwood - thank you for the comments/notes.  Currently, we have only tested with ALB as an external load balancer which is why we may be light for defining other external LBs in the chart.  If you do not use an ALB then it will configure the nginx pod as a k8s load balancer for ingress.  

We made the kubeVersion updates as you suggested.  Never would have figured out the - vs + for the build numbers.  Made, the namespace change for the nginx as well. 

As for the Java Exception, do you have a valid database defined and running?

samuelattwood:
@sbtaylor15 My mistake, I was thinking the chart deployed its own database. I was able to deploy the chart after providing a postgres db.

It looks like it your recent `10.0.29` update you fixed the nginx namespace and kubeVersion issues. So once you update the PR or open a new PR with that version we should be able to merge.

Not a requirement here, but on the topic of the LoadBalancer, from a user perspective it would be useful to be able to use a standard ingress controller rather than having to provision a LoadBalancer resource for the application itself. Just a thought for a future update.

Thanks

PR #401, opened 2022-04-22
HPE CSI Driver for Kubernetes Helm chart 2.1.1

datamattsson:


samuelattwood (approved):
Thanks for the updated PR.

Looks good. Tested on RKE 1.22.7 with Rancher 2.6.4. Approved.

PR #403, opened 2022-04-23
Update Kasten K10 Chart version to 4.5.14

infraq:
Update K10 Chart to 4.5.14


samuelattwood (approved):
Looks good. Tested on RKE 1.22.7 with Rancher 2.6.4. Approved.

PR #408, opened 2022-05-03
Add Pixie to partner charts

aimichelle:
This diff adds Pixie, a CNCF observability tool for Kubernetes, to the partner charts.
It adds the Pixie package, and associated generated chart.


samuelattwood (approved):
Thanks for the submission!

Everything looks good. Tested deployment on  RKE 1.22.9 and RKE2 1.23.6. Confirmed that both clusters appeared in Pixie dash as healthy. Approved.

PR #426, opened 2022-05-13
F5 docs

trinaths:
Updates to F5 Charts documentation.


trinaths:
@samuelattwood  - Please help review and merge this PR. Thank you.

samuelattwood:
@trinaths Apologies for the long delay. We prefer to not modify a chart without iterating the version. Your edits look to be just fine, but it would be appreciated if you could just iterate your `packageVersion` value so we aren't publishing a modified chart under the same version number. Thanks!

PR #433, opened 2022-05-26
NGINX Service Mesh chart v1.4.1

pammecrandall:


samuelattwood (approved):
Tested on RKE2 1.23.6 with Rancher 2.6.5. Approved.

PR #436, opened 2022-06-02
Adding Tetrate Istio Distro

PetrMc:
please review our first submission - if all goes well - we'll continue to support and update versions.

cc @saptak


samuelattwood:
Hey @PetrMc Apologies for the delay.

Just a few things we want to correct before the merge. I'm guessing you pulled some annotations from the Rancher Istio chart, but a few of these are intended for internal use. Please remove `catalog.cattle.io/type` and `catalog.cattle.io/ui-component`.

The only _required_ annotations are `certified`, `display-name`, and `release-name` though you may set the others if you wish. We also generally require setting `kubeVersion` the chart variable, though the annotation serves the same purpose within Rancher.

Other than that, please just be sure to run `make clean` prior to your first commit, as the `packages/tetrate-istio/charts` directory is intended to be ephemeral and not committed.

Let me know if there is anything I can clarify. Thanks

PetrMc:
thanks @samuelattwood - hope that the new commit meets the requirements. Please review at your earliest convinience

samuelattwood:
Hey @PetrMc All the bits we need are present, but it looks like you captured some vscode dot files. While we're at it, we can just do a bit of commit cleanup here and should be all set.

If we start with the current state of your branch:
```bash
export PACKAGE=tetrate-istio
git reset 66eccbd31108e676e09fbb343bbe2474dc9a985d
make clean
git add packages/tetrate-istio
git commit -m "Tetrate Istio package 1.12.6"
git checkout index.yaml
git clean -d -f
make charts
git add index.yaml assets charts
git commit -m "Make charts - Tetrate Istio" #Whatever commit message you like
git push -f origin main-source
```

Thanks

samuelattwood (approved):
Approved.

PR #441, opened 2022-06-09
Fix portworx 2.9.1 versioning for gke

dahuang-purestorage:
In previous PR (https://github.com/rancher/partner-charts/pull/387), we left out changes to add `-0` after versions for gke to be compatible. This PR aims to fix that.


dahuang-purestorage:
@samuelattwood  review welcome

PR #450, opened 2022-06-27
Dynatrace operator v0.7.0

dt-team-kubernetes:


samuelattwood:
Hey @dt-team-kubernetes, thanks for the submission.

Just two things we want to address before merging. First, please add the Rancher annotations to your Chart.yaml, this can be done either in your upstream or just in this repo via the patch mechanism. They should look something like this:
```yaml
annotations:
  catalog.cattle.io/display-name: Dynatrace Operator
  catalog.cattle.io/release-name: dynatrace-operator
```

Second, it looks like the `tests` directory you have in `generated-changes/overlay` are not getting applied to the actual chart, so I am looking in to why that isn't happening. Just to confirm, you do want this directory to be inside the chart despite it not being a part of your chart release gzip? Thanks

dt-team-kubernetes:
Closed in favour of https://github.com/rancher/partner-charts/pull/460

The tests do not need to be applied to the actual chart. 
We only use them to unit test our Helm chart.

PR #454, opened 2022-06-29
Update Dkube: Update installation instructions in app-readme

tijithomas:
@samuelattwood We are trying to update the instructions that are shown on the Chart installation page. We have only updated the readme and the package version. Let us know if we are missing anything required in order to get this updated. Thanks!


samuelattwood:
Hey @tijithomas Sorry for the delay.

The edits you are making are fine, but please avoid removing the old version assets and simply add this as a revision of the chart, that is the intended use of iterating the `packageVersion`. Once a chart is added to the index we try to treat it as immutable to avoid impacting existing deployments. Let me know if I can clarify anything further. Thanks

tijithomas:
@samuelattwood I am afraid I don't follow.
Since the actual changes are only part of app-readme, it made sense not to update the app version, and instead update the packageversion. After updating the package version, 'make charts' resulted in update of the index yaml and the renaming of all other files(with the updated version). Do you think I should skip updating the package version too? Please clarify. Thanks.

samuelattwood:
Hi @tijithomas To clarify, the steps you took were correct, but it looks like your branch is quite a bit behind our upstream main-source, and does not have the commit where the existing dkube chart was merged. As a result, your changes are effectively removing the existing dkube chart version.

You should be good to replicate your steps, but just fetch & merge upstream into your branch beforehand. Thanks

tijithomas:
@samuelattwood Got it. Thanks for clarifying. I would replicate the steps after fetching the upstream branch.

samuelattwood:
Hey @tijithomas I realized right after I replied that my last message was slightly incorrect so I removed it and was typing this reply as you sent yours :slightly_smiling_face:

It looks like your first commit, in addition to updating the `package.yaml` and `app-readme.md` as expected, also is purging `assets/dkube` and `charts/dkube` which is what we want to avoid, we want to add these changes as a revision and not remove the existing version.

The steps you took in iterating the `packageVersion` were correct. I was right in stating that your branch was quite out of date though, so a fetch & merge would also be good. Thanks

tijithomas:
@samuelattwood Got it, thanks!

tijithomas:
@samuelattwood I have replicated the steps after fetching the upstream commits. Let me know if this looks okay.

PR #457, opened 2022-07-05
Loft

thebsdbox:
This adds the loft package in `/package/loft/` and adds all of the autogenerated `/charts/loft/` along with updating the `index.yaml`.


samuelattwood:
Hey @thebsdbox, thanks for the submission.

There are a couple things we do require to  be added before merge. You can review the [workflow](https://github.com/rancher/partner-charts/#workflow) to see these steps documented, but to summarize, please add the Rancher annotations, an app-readme.md file, and set the `kubeVersion` value. All of these may be tracked in this repo using the patch and overlay mechanisms without altering your upstream.

The annotations should look something like this:
```yaml
annotations:
  catalog.cattle.io/display-name: Loft
  catalog.cattle.io/release-name: loft
```

Let me know if I can clarify anything further. Thanks

samuelattwood:
Hey @thebsdbox Apologies for the delay, I was having an issue during deployment and was trying to work out what was going on, since your upstream was deploying just fine.

It looks like your chart is using `.Chart.Version` as a default image tag, however the `packageVersion` modifies the chart version which is causing it to pull a different image than intended. It is generally preferable to use `.Chart.AppVersion` for image tags rather than `.Chart.Version` since the chart version is intended to track changes to the chart itself rather than the underlying application. Helm discusses this a bit in their [docs](https://helm.sh/docs/topics/charts/#the-appversion-field).

Ideally this would be a change in your upstream, however as a temporary solution you could set the `image` value in your `values.yaml` via a patch file in this repo, the same way you are adding the Rancher annotations.

Let me know if you have any questions or if I can clarify anything further. Thanks

samuelattwood:
@thebsdbox I have merged #472 so I will close this out. Thanks :slightly_smiling_face:

PR #461, opened 2022-07-06
Dynatrace operator v0.7.1

dt-team-kubernetes:


samuelattwood:
Hey @dt-team-kubernetes It looks like you just added the annotation to the Chart.yaml in the `charts` directory which means it doesn't make it into the gzip or the index.yaml. Modifications to the chart can be controlled via patch files and tracked in this repo, without having to change your upstream. We also require a `kubeVersion` value to be set in the Chart.yaml, I missed that in my last message. Your docs list 1.21+ as required so you could just set `kubeVersion: '>=1.21-0'`.

If we start with the current state of your branch, this should get us where we need:
```bash
export PACKAGE=dynatrace-operator
git reset 12f1d115b28111c62efea86d4e737c338a92d84c
rm -rf packages/dynatrace-operator/generated-changes/overlay/tests/
make prepare

##Edit packages/dynatrace-operator/charts/Chart.yaml to add annotations and set kubeVersion

make patch
make clean
git add packages/dynatrace-operator
git commit -m "Make patch"
git checkout index.yaml
git clean -d -f 
make charts
git add index.yaml assets charts
git commit -m "Make charts"
git push -f origin dynatrace-operator-v0.7.1
```
In the above I removed your `tests` directory since it isn't serving any function here, but it isn't hurting anything so it's not a huge deal either way. Let me know if I can clarify anything further. Other than the `Chart.yaml` edit, you can basically copy and paste the above if you'd like to. Thanks

dt-team-kubernetes:
Closed in favour of https://github.com/rancher/partner-charts/pull/467

PR #462, opened 2022-07-11
Komodor

nagauker:
Thanks


samuelattwood:
Hey @nagauker Thanks for the submission, everything looks good but one suggestion I'd make is to set your `kubeVersion` to `kubeVersion: '1.16-0 - 1.21-0'`, the `-0` suffix avoids some issues with EKS & GKE since their versioning causes Helm to mark them pre-release. This can result in Helm considering them 'incompatible' and Rancher hiding the chart for that cluster.

nagauker:
Hey, @samuelattwood  Thank you for your comment, I changed as you suggested! :)

samuelattwood:
Hey @nagauker, your fix looks good but we need make sure the index and asset gzip get regenerated. If you do the following, starting with your branch's current state, we should be all set:

```
export PACKAGE=komodor
git reset f53135e6f7cbf973a9bc24074b3ed6acee938e78
git add packages/komodor assets/logos/komodor.png
git commit -m "Komodor kick"
git checkout index.yaml
git clean -d -f
make charts
git add assets charts index.yaml
git commit -m "Make charts"
git push -f origin komodor
```

This just resets to right before your first commit, recreates it with the kubeVersion fix included, then regenerates the chart and index. This cleans up the commit history a bit as well. Thanks

nagauker:
Hi, @samuelattwood 😄.
I really appreciate the guidance and help 🥇 , I did what you suggested

samuelattwood (approved):
Thanks for the update. Looks good, approved.

PR #464, opened 2022-07-12
(NGINX) Update NGINX Ingress Controller to App Version 2.3.0 - Helm Chart 0.14.0

ciarams87:
PR to update the catalogue with our latest release.

Tested in Rancher v2.6.5
Screenshot 2022-07-12 at 14 08 48


samuelattwood:
Hey @ciarams87 Quick questions about your `kubeVersion`. It looks like you're changing it here from the upstream of `'>= 1.19.0-0'` to `'1.19 - 1.24'`. That's fine if you want to make that change, but I would recommend you keep the `-0` suffix, it will avoid some issues with GKE and EKS versioning and SemVer2, so I would set it as `'1.19-0 - 1.24-0'`, otherwise it might get marked as incompatible and hidden for GKE and EKS clusters.

samuelattwood (approved):
Approved.

PR #467, opened 2022-07-13
Dynatrace operator v0.7.2

dt-team-kubernetes:


samuelattwood:
@chrismuellner Apologies for the very long delay here. I have been working on finalizing a new config option to automate chart updates.

I have just submitted a PR (#507) on your behalf to hide the old 'oneagent' operator and add the Dynatrace Operator using this new configuration. With this merged, your upstream helm repo will be checked daily for updates, and if present they will be pulled in automatically. Since you already publish the app-readme.md and questions.yaml in your upstream, you should no longer need to submit PRs to this repo unless your Helm repo address were to change.

Let me know if you have any questions, and with your confirmation I will proceed with merging the change. Thanks

chrismuellner:
@samuelattwood thanks for the update, sounds good to me!

I'm assuming that updates to either `app-readme.yaml` or `questions.yaml` will be read when we release a new chart version, or will that require a PR from our side since it's not in the folder you are watching?

samuelattwood:
@chrismuellner The Helm repo index that will be watched is [this file](https://github.com/Dynatrace/dynatrace-operator/blob/master/config/helm/repos/stable/index.yaml), and your chart gzip releases do currently contain an `app-readme.md` and `questions.yaml`, so as long as that continue to be the case they will be updated just like the rest of the chart.

If you later decide you wish to instead track those files in this repo, you still may do so, you would just submit a PR to store them in `packages/dynatrace/dynatrace-operator/overlay` and they will be added to any future chart updates.

I will proceed with merging #507 🙂

PR #475, opened 2022-07-29
Nutanix csi snapshot 6.0.1

tuxtof:


samuelattwood (approved):
It looks like a `Chart.yaml.orig` got captured in the commit. It won't impact anything, so I'll go ahead and approve so this can be merged along with your other chart update. It would be good to remove in a future commit.

PR #488, opened 2022-08-12
Typo

vrevelas:


samuelattwood:
This is a typo in a fairly old version of Neuvector, that has been fixed for some time in newer versions. I'm going to close this out rather than regenerate the chart files for the older version just for this.

PR #492, opened 2022-08-15
External secrets operator 0.5.9

CharlKlein:
Helllos,

Please accept this PR for the Updates to increment External-Secrets-Operator to v0.5.9 :)

Sharing some awesome news, ESO has been accepted in to the CNCF Sandbox 🥳 

Tested on Rancher 2.6.4
```
NAME                                                             READY   STATUS    RESTARTS   AGE
pod/external-secrets-operator-5f5b555cdf-26z9k                   1/1     Running   0          7m20s
pod/external-secrets-operator-cert-controller-5649dc8b44-lxm74   1/1     Running   0          7m20s
pod/external-secrets-operator-webhook-5cf4c657df-8jjc9           1/1     Running   0          7m20s

NAME                                        TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)   AGE
service/external-secrets-operator-webhook   ClusterIP   10.43.111.16   <none>        443/TCP   7m20s
service/kubernetes                          ClusterIP   10.43.0.1      <none>        443/TCP   10m

NAME                                                        READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/external-secrets-operator                   1/1     1            1           7m20s
deployment.apps/external-secrets-operator-cert-controller   1/1     1            1           7m20s
deployment.apps/external-secrets-operator-webhook           1/1     1            1           7m20s

NAME                                                                   DESIRED   CURRENT   READY   AGE
replicaset.apps/external-secrets-operator-5f5b555cdf                   1         1         1       7m20s
replicaset.apps/external-secrets-operator-cert-controller-5649dc8b44   1         1         1       7m20s
replicaset.apps/external-secrets-operator-webhook-5cf4c657df           1         1         1       7m20s
```

PR #503, opened 2022-08-22
Migrating 'external-secrets' config

samuelattwood:


samuelattwood:
#492

Hey @CharlKlein I have migrated your chart over to the newer config.  To overview the changes, the `packages` and `charts` directories are now organized as `vendor/chart` in this case being `external-secrets/external-secrets`. The important new file is the `upstream.yaml` which defines where the CI should check for chart updates. All options for this file are listed [here](https://github.com/samuelattwood/partner-charts-ci#configuration-file), but as it stands you shouldn't need to change anything in this file.

With this configuration, you should only need to submit a PR if you want to modify files in the `packages/external-secrets/external-secrets/overlay` directory which are your `app-readme.md` and `questions.yaml`. The upstream will be checked daily for new releases.

For the time being we are keeping the `PackageVersion` variable to ensure new releases are correctly listed as updates, but once a version `0.6.0` releases we will drop that and match your upstream chart version.

Let me know if you have any questions. Thanks

alex-isv (approved):

PR #485, opened 2022-08-07
Update px essentials 2.10

dahuang-purestorage:

PR #484, opened 2022-08-07
Update px 2.10.3

dahuang-purestorage:

PR #504, opened 2022-08-23
Migrating 'portworx' and 'portworx-essentials' config

samuelattwood:


samuelattwood:
#484 #485

Hey @dahuang-purestorage I have migrated your charts over to the newer config. To overview the changes, the packages and charts directories are now organized as `vendor/chart` in this case being `portworx/portworx` and `portworx/portworx-essentials`. The important new file is the `upstream.yaml` which defines where the CI should check for chart updates. All options for this file are listed [here](https://github.com/samuelattwood/partner-charts-ci#configuration-file), but as it stands you shouldn't need to change anything in these files. The upstream will be checked daily for chart updates and they will be pulled in automatically.

Let me know if you have any questions. Thanks

alex-isv (approved):

PR #513, opened 2022-09-06
Add RKE2 in the k3s containerd runtime section

abonillabeeche:
For visibility/ease of use, including RKE2 on the k3s runtime section

Signed-off-by: Alejandro Bonilla <abonilla@suse.com>


samuelattwood:
@abonillabeeche We want to make sure the chart files remain consistent across the assets and charts directories. I have made these same edits but updated the other relevant files accordingly in #514. I will close this out in favor of that PR.

PR #614, opened 2022-12-16
Charts CI

redhatrises:
```
Updated:
  crowdstrike/falcon-sensor:
    - 1.18.1
```


samuelattwood:
Hey @redhatrises Your PR looks good and you followed the updated process correctly.

I went ahead and made #615 for just a couple things. When migrating from a chart that used the prior method, there is some slight additional cleanup we typically do to avoid having versions of charts stored in separate directories.

I also did not include the `kubeVersion` and `icon` in the `upstream.yaml` because the upstream chart already sets these values so you don't need to overwrite them here, unless there is a specific reason you wanted to do so.

If that looks good to you, we could proceed with that PR.

redhatrises:
> Hey @redhatrises Your PR looks good and you followed the updated process correctly.
> 
> 
> 
> I went ahead and made #615 for just a couple things. When migrating from a chart that used the prior method, there is some slight additional cleanup we typically do to avoid having versions of charts stored in separate directories.
> 
> 
> 
> I also did not include the `kubeVersion` and `icon` in the `upstream.yaml` because the upstream chart already sets these values so you don't need to overwrite them here, unless there is a specific reason you wanted to do so.
> 
> 
> 
> If that looks good to you, we could proceed with that PR.

@samuelattwood sounds good. Let's do that.

alex-isv (approved):

PR #616, opened 2022-12-19
add Percona Operator for MySQL based on Percona XtraDB Cluster

nmarukovich:
Signed-off-by: Natalia Marukovich <natalia.marukovich@percona.com>

### DIFF
--- a/packages/percona/pxc-db/upstream.yaml
+++ b/packages/percona/pxc-db/upstream.yaml
@@ -1,0 +1,7 @@
+HelmRepo: https://percona.github.io/percona-helm-charts
+HelmChart: pxc-db
+Vendor: Percona
+DisplayName: Percona Operator for MySQL based on Percona XtraDB Cluster
+ChartMetadata:
+  icon: https://raw.githubusercontent.com/percona/percona-xtradb-cluster-operator/main/operator.png
+  kubeVersion: '>=1.21-0'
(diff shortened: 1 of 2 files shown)


nmarukovich:
@samuelattwood hi! Could you have alook on this PR?

nflondo:
Hi @nmarukovich Do you want both charts to be named the same in the Rancher Catalog? "Percona Operator for MySQL based on Percona XtraDB Cluster"  Also, the name seems a bit long and it may get cut off in the UI.

nmarukovich:
> Hi @nmarukovich We don't allow two charts to have the same name in the Rancher Catalog: "Percona Operator for MySQL based on Percona XtraDB Cluster" Also, the name seems a bit long and it may get cut off in the UI.

ok. So I need create separate catalog for new operator?

nflondo:
No, you only need to change the "DisplayName:" in your upstream.yaml files to make the name of each chart unique (and shorter preferably).  Right now, both charts have the same long name and it looks like this in the Rancher catalog:
percona-new

nmarukovich:
> No, you only need to change the "DisplayName:" in your upstream.yaml files to make the name of each chart unique (and shorter preferably). Right now, both charts have the same long name and it looks like this in Rancher Apps: percona-new

O! Thanks! Fixed! 
Unfortunately I can't make the name of operator shorter and we need to use 
Percona Operator For MySQL based on Percona XtraDB Cluster 
because we also have operator for mysql.

nmarukovich:
@nflondo could you please recheck it?

nflondo (approved):
Approved

PR #683, opened 2023-03-10
Metal3 helm charts

chajain:
Adding Metal3 Helm charts 
- media
- ironic
- baremetal-operator
- powerdns
- metal3-deploy


nflondo:
Closing this PR with Jeremy Moffitt's approval: "Talked direction with Tim this morning and for the moment we agreed it doesn't make sense to pursue the partner-charts approach. I'm not sure if we want to close Charu's PR or mark it as "draft" or whatever the equivalent may be for that repo, we'll want to keep the work around as a reference short term... 
Thanks,
Jeremy
________________________________________
"

PR #787, opened 2023-06-09
Kubecost fix eks option

linhlam-kc:
Add a workaround solution to fix EKS option default value not using the correct image URI


nflondo (approved):
Approved.

PR #797, opened 2023-06-15
Update Chronicle helm chart upstream source

alexhq:
This updates the source for BTP's Chronicle helm chat from using the chart source git repo to use our published helm repository. 
 
One outstanding question before this is ready is why the location of the dependancies has changed in index.yaml

### DIFF
--- a/packages/btp/chronicle/upstream.yaml
+++ b/packages/btp/chronicle/upstream.yaml
@@ -1,6 +1,4 @@
-GitRepo: https://github.com/btpworks/chronicle.git
-GitBranch: main
-GitSubdirectory: charts/chronicle
-RemoteDependencies: true
+HelmRepo: https://btp-charts-stable.s3.amazonaws.com/charts
+HelmChart: chronicle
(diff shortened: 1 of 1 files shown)


nflondo (dismissed):

nflondo:
Hi @alex, we haven't made changes to our workflow with regards to location of dependencies. I believe you may be referring to dependencies' repository being either URL or local file (./charts/<chart>) if so, we use the URL when upstream.yaml is configured to fetch from github repo with flag RemoteDependencies: true but when upstream.yaml is configured to fetch from Helmrepo we just unpack the dependencies and put them in ./charts/<chart>

alexhq:
Hi @nflondo Thank you, that makes sense, as we're swapping to fetch from a Helm repo in this PR, we should update the previous versions of Chronicle so that they use ./charts/<chart>. 

I've updated the PR as requested to only include the change in `packages/btp/chronicle/upstream.yaml`

nflondo (approved):
Approved

PR #942, opened 2023-11-08
 Address warning message in the CI pipelines

### LINKED ISSUE #939: Address warning message in the CI pipelines
We want to address [this warning message](https://github.com/rancher/partner-charts/actions/runs/6785478970) related to the use of a very old version of Node.

Github [announcement](https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/) with more details.

#### Acceptance criteria

- [ ] The warning message has disappeared
- [ ] All pipeline are reviewed

atrendafilov:
### Issue

Github CI pipelines action is completed successfully, but with warning about outdated Node.js version

### Solution

Upgrade default Node.js version in the Ubuntu image from 12 to 16

### Testing done

Successfully tested

Fix #939

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -12,10 +12,7 @@ jobs:
       CI_TOOLS_URL: https://github.com/samuelattwood/partner-charts-ci/releases/latest/download/partner-charts-ci-linux-amd64
 
     steps:
-      - uses: actions/setup-node@v2
-        with:
-           node-version: '16.x'
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           fetch-depth: 0
 
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -12,6 +12,9 @@ jobs:
       CI_TOOLS_URL: https://github.com/samuelattwood/partner-charts-ci/releases/latest/download/partner-charts-ci-linux-amd64
 
     steps:
+      - uses: actions/setup-node@v2
+        with:
+           node-version: '16.x'
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
@@ -36,5 +39,5 @@ jobs:
         run: git rebase origin/main-source
 
       - name: Validate 
-        if: "!contains(github.event.pull_request.title, '[modified charts]')"
+        if: '!contains(github.event.pull_request.title, "[modified charts]")'
         run: partner-charts-ci validate
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -39,5 +39,5 @@ jobs:
         run: git rebase origin/main-source
 
       - name: Validate 
-        if: '!contains(github.event.pull_request.title, "[modified charts]")'
+        if: "!contains(github.event.pull_request.title, '[modified charts]')"
         run: partner-charts-ci validate


recena:
@atrendafilov In this PR you are updating one pipeline but there are [three pipelines](https://github.com/rancher/partner-charts/tree/main-source/.github/workflows). Is there any reason not to update the rest?

.github/workflows/pull-request.yaml
```diff
@@ -12,6 +12,9 @@ jobs:
       CI_TOOLS_URL: https://github.com/samuelattwood/partner-charts-ci/releases/latest/download/partner-charts-ci-linux-amd64
 
     steps:
+      - uses: actions/setup-node@v2
+        with:
+           node-version: '16.x'
```
recena, line 17: Why not `node-version: 18`?

.github/workflows/pull-request.yaml
```diff
@@ -12,6 +12,9 @@ jobs:
       CI_TOOLS_URL: https://github.com/samuelattwood/partner-charts-ci/releases/latest/download/partner-charts-ci-linux-amd64
 
     steps:
+      - uses: actions/setup-node@v2
```
recena, line 15: Why not the latest version?

.github/workflows/pull-request.yaml
```diff
@@ -12,6 +12,9 @@ jobs:
...
+        with:
+           node-version: '16.x'
       - uses: actions/checkout@v2
```
recena, line 18: I think the fix is to update this action to [the latest version](https://github.com/marketplace/actions/checkout).

recena: I'm almost sure this fix the warning message. More references https://github.com/actions/checkout/issues/1047#issuecomment-1355170319

recena: `actions/checkout@v2` [uses Node12](https://github.com/actions/checkout/blob/v2.0.0/action.yml#L33).

atrendafilov:
> @atrendafilov In this PR you are updating one pipeline but there are [three pipelines](https://github.com/rancher/partner-charts/tree/main-source/.github/workflows). Is there any reason not to update the other?

Because in the other pipelines we use `actions/checkout@v3 ` which mitigates this issue.

.github/workflows/pull-request.yaml
atrendafilov: version 16 is sufficient enough and I didn't want to make a huge leap in versions.

.github/workflows/pull-request.yaml
atrendafilov: same as the above comment about the versions

.github/workflows/pull-request.yaml
atrendafilov: This could be a possible fix, but I'm not familiar with the reason behind the use of `actions/checkout@v2` in this pipeline, so just to be on the save side I have opted out for this solution.

recena: @atrendafilov This is the fix according to the documentation and references. In fact, we don't need to add a new step.

.github/workflows/pull-request.yaml
recena: Node 16 is EOL. https://endoflife.date/nodejs

.github/workflows/pull-request.yaml
recena: Node 16 is EOL. https://endoflife.date/nodejs

.github/workflows/pull-request.yaml
recena: I don't see the value of not using the latest version of the official Github Actions.

recena (changes requested):
Go through updating `actions/checkout@v2`

recena (approved):

recena:
Merging to avoid conflicts.

PR #970, opened 2024-01-19
Submitting containers/nri-plugins

fmuyassarov:
Submitting the following [containers/nri-plugins](https://github.com/containers/nri-plugins)
- nri-memory-qos
- nri-memtierd
- nri-resource-policy-balloons
- nri-resource-policy-template
- nri-resource-policy-topology-aware
- nri-sgx-epc

### DIFF
--- a/packages/containers/nri-memory-qos/upstream.yaml
+++ b/packages/containers/nri-memory-qos/upstream.yaml
@@ -1,0 +1,6 @@
+ArtifactHubRepo: nri-plugins
+ArtifactHubPackage: nri-memory-qos
+Vendor: containers
+DisplayName: Memory-QoS NRI plugin
+ChartMetadata:
+  kubeVersion: '>=1.24-0'
--- a/packages/containers/nri-memtierd/upstream.yaml
+++ b/packages/containers/nri-memtierd/upstream.yaml
@@ -1,0 +1,6 @@
+ArtifactHubRepo: nri-plugins
+ArtifactHubPackage: nri-memtierd
+Vendor: containers
+DisplayName: Memtierd NRI plugin
+ChartMetadata:
+  kubeVersion: '>=1.24-0'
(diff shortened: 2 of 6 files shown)


packages/containers/nri-memory-qos/upstream.yaml
```diff
@@ -0,0 +1,6 @@
+ArtifactHubRepo: nri-plugins
+ArtifactHubPackage: nri-memory-qos
+Vendor: containers
```
recena, line 3: Is this correct?

fmuyassarov: If you mean the `Vendor:`, then I was not totally sure if I had to write `nri-plugins` (which is the GitHub project name) or the GitHub org (`containers`) that is hosting `nri-plugins` project.

And I didn't find something that would be vendor neutral. nri-plugins isn't a particular vendor owner project but community maintained project. 

Do you think I should replace it by nri-plugins instead?

recena: `Vendo` means the company responsible for this software.

recena: I assume you are adding this software because you company has a partner relationship with SUSE / Rancher. Right?

recena:
My suggestion is to add this repository (https://containers.github.io/nri-plugins) to Rancher Manager. In this way, you can access to the official Helm Charts.

packages/containers/nri-memory-qos/upstream.yaml
fmuyassarov: Actually this project is not tied to a company.

fmuyassarov:
> @fmuyassarov Let me close this PR. Ok?

Yes, we can close it. Sorry for the noise.

PR #983, opened 2024-03-06
Automate updates of existing helm charts

adamkpickering:
`rancher/partner-charts` adds a Rancher-specific layer to a variety of upstream charts. These charts are updated frequently, and the Rancher versions of these charts don't get updated unless we go through a manual process to do so. This process has most recently been done by @marcosbc. We want to automate this process so that updates happen more frequently (i.e. daily) and don't sap our attention.

This PR introduces two new workflows:

1. **update-main-source.yml**: runs `bin/partner-charts-ci auto` daily in order to update the `main-source` branch
2. **update-main.yml**: syncs `assets/` and `index.yaml` from the `main-source` to the `main` branch when changes are pushed to the `main-source` branch

There are a lot of things with this repo that I'd like to change. However, this PR is just about automating what we have. We can make further improvements later.


nicholasSUSE (changes requested):
Sorry @adamkpickering, I have no idea what this is about, could you provide a description?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,32 @@
...
+        scripts/pull-ci-scripts
+        bin/partner-charts-ci auto
+        git push origin main-source
+
+    - name: Update main branch
+      run: |
+        # checkout action only fetches main-source, so we need to fetch main branch also
+        git fetch origin main --depth 1
+        git checkout main
+        git checkout main-source -- index.yaml assets
```
marcosbc, line 30: The original action also removes the `assets` folder and the `index.yaml` file before executing the checkout. Have you tested that it works?

marcosbc (changes requested):
We should also remove the `.github/workflows/sync-fork.yml` file given that it will not be applicable anymore.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - name: Configure git
+      run: |
+        git config --global user.name "github-actions[bot]"
+        git config --global user.email "github-actions[bot]@users.noreply.github.com"
+
+    - name: Update main-source branch
+      run: |
+        scripts/pull-ci-scripts
+        bin/partner-charts-ci auto
+        git push origin main-source
```
marcosbc, line 23: Instead of pushing directly to the branches, wouldn't it be better to create PRs for each individual change, with auto-merge enabled?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+    - name: Configure git
+      run: |
+        git config --global user.name "github-actions[bot]"
+        git config --global user.email "github-actions[bot]@users.noreply.github.com"
+
+    - name: Update main-source branch
+      run: |
+        scripts/pull-ci-scripts
+        bin/partner-charts-ci auto
```
ruospalo, line 22: We are skipping `bin/partner-charts-ci validate`  [step](https://github.com/rancher/partner-charts-ci#5-run-a-validation)  to ensure released assets are not being modified, is this step neede now?

.github/workflows/release.yaml
adamkpickering: What do you mean by "each individual change"? Are you suggesting that this automation should create a PR that is merged automatically every time it runs? If so, I can't think of a benefit to doing this - can you please elaborate?

If this isn't what you meant, can you clarify?

marcosbc: By creating a PR (even with auto merge enabled), it forces an execution of the CI scripts to validate the change. Currently, this executes `partner-charts-ci validate` (see [.github/workflows/pull-request.yml](https://github.com/rancher/partner-charts/blob/main-source/.github/workflows/pull-request.yaml)). If something went wrong, we would have the ability to stop the change from getting merged.

.github/workflows/release.yaml
adamkpickering: The purpose of this automation is to release updated assets, i.e. modify them. So we don't want to ensure that they are not modified. Or maybe you know something I don't?

ruospalo: Thank you Adam. AFAIK, It ensures that we don't modify the already released assets (not the one that we are updating). In any case, is just to be sure that we are skipping this step consciously.

.github/workflows/release.yaml
adamkpickering: I think I should have left that in. I'll add it back.

marcosbc (dismissed):
LGTM!

.github/workflows/update-main-source.yml
```diff
@@ -0,0 +1,41 @@
...
+    - uses: actions/checkout@v4
+
+    - name: Configure git
+      run: |
+        git config --global user.name "github-actions[bot]"
+        git config --global user.email "github-actions[bot]@users.noreply.github.com"
+
+    - name: Update main-source branch
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
recena, line 21: @adamkpickering 🤔 Which one?

marcosbc (approved):

.github/workflows/update-main-source.yml
```diff
@@ -0,0 +1,41 @@
...
+        scripts/pull-ci-scripts
+        BRANCH="auto-update/$(date '+%s')"
+        git checkout -b "$BRANCH"
+        bin/partner-charts-ci auto
+
+        # exit if there are no changes
+        git diff --quiet main-source "$BRANCH" && exit 0
+
+        # close all existing PRs from branches starting with "auto-update"
+        gh pr --repo "$GITHUB_REPOSITORY" list --search 'head:auto-update' --json 'headRefName' --jq '.[] | join("\n")' | \
```
nicholasSUSE, line 34: One possible issue is that you should always make sure that this Ubuntu latest image has JQ installed; otherwise, it will break. 
My Ubuntu did not come with it installed.

nicholasSUSE (approved):

.github/workflows/update-main-source.yml
adamkpickering: I don't think this command requires `jq` - it's just an option. Looking at the dependencies of the `gh` package, only `git` is listed. So they must build `jq`, or its syntax, into `gh` somehow?

PR #997, opened 2024-04-03
Automate updates of existing helm charts

adamkpickering:
This is a second try at #983. For context see https://github.com/rancherlabs/eio/issues/2112 and the JIRA issue.

A quick summary: an installation access token is used to authenticate as a github app when pushing code. This github app is given an exception to the branch protection rules for this repo, which normally prevent pushes directly to the `main-source` and `main` branches.

I tried to get automated updates working with PRs, but github does not allow this without a human in the loop. This is why these workflows now simply create commits on `main-source` and `main` and push them.

I tested the parts of this that I could, but github/our setup makes full testing on a fork impossible. I am confident that these changes won't screw this repo up, but I am not confident that it will work on the first try.

### DIFF
--- a/.github/workflows/update-main-source.yml
+++ b/.github/workflows/update-main-source.yml
@@ -16,26 +16,29 @@ jobs:
         git config --global user.name "github-actions[bot]"
         git config --global user.email "github-actions[bot]@users.noreply.github.com"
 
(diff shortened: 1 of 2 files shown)


lucasmlp (approved):
LGTM 🚀

PR #1011, opened 2024-04-17
Remove charts managed by Koor

galexrt:
Koor Technologies, Inc. is in the process of shutting down, see https://koor.tech/.

Please accept this PR to remove the Koor's `koor-operator` chart.


adamkpickering (changes requested):
Thank you for the PR! I see that a few more things need to be removed:
- `packages/koor-tech/`
- `assets/koor-tech`
- References to charts managed by Koor in `index.yaml`

galexrt:
Thanks! I have updated the PR to remove the other files and update the `index.yaml` file. PTAL

adamkpickering (approved):
LGTM. Thanks again :smile:

adamkpickering:
FYI I'm going to hold off on merging this, because we are currently working on some automation and I want to avoid causing any problems with that work.

PR #1016, opened 2024-05-08
Adding icons to override

nicholasSUSE:
### Problem
Partner Charts has a lot of Charts with several logos being fetched online. 
This causes some problems for AirGap Environments. 

[Related partner-charts-ci PR](https://github.com/rancher/partner-charts-ci/pull/3)

This PR is the result of the execution of the `download-icons` automation added in the above PR. 
After this is merged, we must execute through the CI the `override-icons` automation from the `partner-charts-ci`

### Solution 
Adding the icon's local files to enable access by the Rancher Manager through the local system files to avoid fetching the internet. 

Also, editing/adding upstream.yaml files in packages folder so we can override these file paths on the index.yaml through CI automation later.


adamkpickering (approved):

recena:
Before merging, we should cut a release here https://github.com/rancher/partner-charts-ci/releases

PR #1019, opened 2024-05-15
Remove `argo/argo-cd` package

adamkpickering:


recena (approved):

arley-wilches:
hi guys, what was the reason for deleting Argo?
was it for something special to take into account? I used that chart several times.

Thank you.

arley-wilches:
hi @troytop 
thank you for your response, I have an additional question, I used to install apps from the App section and I used custom repositories too, so, my question is, can I install the app from the Rancher Prime customers link in my on-prem, or is it only for a cloud-only solution?

do you get what I want to ask?

Thank you!

troytop:
Excellent question, which I probably should have anticipated! The Rancher Prime Application Collection is an authenticated OCI Helm repository which Rancher's Apps interface does not yet support. You need to use the helm cli to install those charts right now, but support for custom OCI Helm repositories is coming soon in Rancher 2.9.

arley-wilches:
hi @troytop I got the point

and again, thank you for your time and reply.

PR #1023, opened 2024-06-06
Update `gopaddle/gopaddle` version in place

adamkpickering:
It appears that the upstream chart for `gopaddle/gopaddle` was changed without changing the version number. Because `partner-charts-ci` only looks for new versions, and not changes *within* a version, it did not propagate this change to the partner charts repo. Specifically, this was noticed because the `values.yaml` file was referring to an image tag that does not exist.

This PR manually fixes this problem.

Note: the checks are failing, but this is expected. This is because we have a validation that checks that existing charts are not being modified. This PR does exactly that, but only because upstream did it.


recena (approved):

recena:
@adamkpickering But the pipeline is failing.

index.yaml
```diff
@@ -18530,11 +18530,11 @@ entries:
...
     apiVersion: v2
     appVersion: "5"
-    created: "2024-06-05T00:55:25.01459904Z"
+    created: "2024-06-06T15:15:11.685530761-06:00"
     description: Kubernetes IDE with AI Co-pilot
-    digest: 4f56e39c70605bdeac453f012e472df3652b245a9d9073e4d902f31cd7bae8ff
+    digest: 9791c680e52099b267dfadb74275c7a5de0b866dbfdbadb656ac6c254acd8aae
     home: https://gopaddle.io
-    icon: file://assets/icons/gopaddle.png
+    icon: https://gopaddle-marketing.s3.ap-southeast-2.amazonaws.com/gopaddle.png
```
recena, line 18537: @adamkpickering We discussed the use of external resources for `icon`. This impacts air-gapped environments.

adamkpickering: Hmm that shouldn't have happened. I will investigate

adamkpickering: Looks like this was resolved by the automated update that happens overnight.

PR #1035, opened 2024-06-21
Submit Airlock Microgateway and Airlock Microgateway CNI

tyxeron:
Hi :wave: 

These two commits add the `Airlock Microgatway`  and  `Airlock Microgateway CNI` Helm charts.

Thanks for reviewing

GitHub Repo: https://github.com/airlock/microgateway

### DIFF
--- a/packages/airlock/microgateway-cni/overlay/questions.yml
+++ b/packages/airlock/microgateway-cni/overlay/questions.yml
@@ -1,0 +1,18 @@
+questions:
+  - variable: config.cniNetDir
+    required: true
+    type: string
+    label: CNI Network Configuration Directory
+    group: "CNI Settings"
+    description: "Directory where the CNI config files reside on the host. This value depends on the kubernetes distribution and interface CNI Provider used. It can be fetched by running `crictl info -o go-template --template '{{.config.cni.confDir}}'` on your kubernetes host."
+  - variable: config.cniBinDir
+    required: true
+    type: string
+    label: CNI Plugin Binaries Directory
+    group: "CNI Settings"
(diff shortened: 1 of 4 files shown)


recena:
📔 This contribution will be reviewed soon.

troytop:
@tyxeron please see the email I sent regarding the SUSE Ready partner application. We'll review and merge the PR after that process is complete (and I'll update the README so this is clearer for others).

adamkpickering:
The partner charts build tool (`partner-charts-ci`) does not provide a way to set kubernetes distribution specific Helm values. At the time that an upstream chart is processed for inclusion into the partner charts repo, `partner-charts-ci` does not have any info about what environment the chart will be installed into. To the best of my knowledge, your options are:

- Provide different charts for different environments. I don't recommend this, since this is not the typical way of doing things and it could confuse users.
- Have the users configure your application for the environment they are installing it into. [`questions.yml`](https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/helm-charts-in-rancher/create-apps#additional-files-for-rancher-charts) may come in handy for this - you would add it as an overlay file. Alternatively the user can configure the application via `Values` as is usually done when installing a chart via `helm install`. The Rancher UI provides a way to do this without having to use the `helm` CLI tool.

I hope that helps :smile:

tyxeron:
@adamkpickering Thanks for your reply :) I already have a questions.yml file so I hope that should be sufficient for the user to install it.

adamkpickering (commented):
Looks good to me. However we are in the middle of a migration, so I am going to hold off on approval. We should have the migration completed in the next day or two.

adamkpickering (approved):

PR #1050, opened 2024-07-10
Submitting cerbos/cerbos

alexolivier:

PR #1051, opened 2024-07-12
Fix icon URL for some (not all) charts

adamkpickering:
In https://github.com/rancher/partner-charts-ci/pull/22 I modified the `download-icons` subcommand of `partner-charts-ci` (and also renamed it to `ensure-icons`). It now does more. This PR is the result of running `partner-charts-ci ensure-icons` on this repo - it downloads icons for more charts than we could get before.

If you're curious as to why we couldn't get these icons before, the answer is technical debt. There was a function called `populatePackages()` that was the only way to list packages. But this function did more than that: it also checked whether there were any new upstream versions for a package. For some packages (i.e. the ones for which icons are added in this PR), there were errors when running `populatePackages()`, so they were skipped.

The reason I am able to add icons for them here is because I have split `populatePackages()` by function: `listPackageWrappers()` now lists packages, and `PackageWrapper.Populate()` checks upstream for new versions. `partner-charts-ci ensure-icons` only needs to call `listPackageWrappers()`, so we can avoid the errors on `PackageWrapper.Populate()` for these packages.


recena:
What about the new ones like [this](https://github.com/rancher/partner-charts/pull/1050/files#diff-f41e7b9e36fe3cbdb52e14a4b3a331f62db067187dedaf79523c6dd7dce7349eR6)?

adamkpickering:
After PRs that add new packages are merged, running `partner-charts-ci auto` will ensure the icons are there in addition to the other things that it does.

diogoasouza (approved):

recena:
`assets/icons/cf-runtime.jpg` looks a wrong file.

adamkpickering:
Good catch @recena. It should be fixed now.

recena (approved):

PR #1057, opened 2024-07-16
Fix corrupted icons in `assets/icons`

adamkpickering:
Because of a bug in `partner-charts-ci`, some icons were corrupted when they were originally downloaded in db77b3ab80dce8952913900a28f7db072c8770ad. This PR fixes those icons. Now, every icon in `assets/icons` should be a valid image file - I have verified that this is the case.

Note that the `partner-charts-ci validate` command that CI runs will fail here. This is expected and not a problem.


adamkpickering:
Please see the message from the main comment:
> Note that the partner-charts-ci validate command that CI runs will fail here. This is expected and not a problem.

recena:
> Please see the message from the main comment:
> 
> > Note that the partner-charts-ci validate command that CI runs will fail here. This is expected and not a problem.

Why is expected?

diogoasouza (approved):

PR #1067, opened 2024-09-30
Deprecate `jenkins/jenkins` package

adamkpickering:
The `jenkins/jenkins` package does not meet the [criteria for inclusion](https://github.com/rancher/partner-charts?tab=readme-ov-file#suse-ready-partnership) into the partner charts repository. Specifically, it is not provided by a SUSE "Ready" Verified partner. Accordingly I am deprecating it according to the [deprecation and removal policy](https://github.com/rancher/partner-charts?tab=readme-ov-file#deprecation-and-removal-policy).

Users should get Jenkins directly from the official repository as described [here](https://www.jenkins.io/doc/book/installing/kubernetes/#install-jenkins-with-helm-v3).

It is expected that CI fails for this PR. This is because CI checks that we do not modify released .tgz files. This is an old design that no longer applies. A work item for overhauling this is present and should be worked on in the near future. Please disregard the CI failure.

A note about the changes in this PR:
- The chart version tarball files must be modified because we are adding `deprecated: true` to the `Chart.yaml` file in each chart version.
- The fields in the `upstream.yaml` file are reordered because `partner-charts-ci deprecate` does not retain the order of fields when it loads this file. The YAML library that `partner-charts-ci` uses to write the file orders fields alphabetically, leading to the change.


recena:
@adamkpickering Can you address the conflict?

adamkpickering:
@recena done.

recena (approved):

diogoasouza:
Ci is failing @adamkpickering

adamkpickering:
@diogoasouza it is expected that CI will fail, since we are modifying existing chart versions, and `partner-charts-ci validate` fails when that is the case. This is not consistent with the design of other `partner-charts-ci` subcommands, and I plan on addressing this in the near future.

PR #1073, opened 2024-10-09
Add `linux-polska/neuvector-security`

qwiatu-lp:
Chart requires  SUSE Neuvector solution to work. The add-on uses following Neuvectr's CRDs:
* nvclustersecurityrules.neuvector.com
* nvdlpsecurityrules.neuvector.com


troytop:
This repository is for software that ISV partners provide to the Rancher community, it cannot be used for special configurations of charts that are maintained by other vendors, in this case SUSE. 

You can make this chart (and the ones in the previous PR) available to Rancher users by documenting how to add the sourcemation.github.io/charts repository in the Rancher UI.

PR #1075, opened 2024-10-09
Fix `Hidden` property of `linux-polska/ezd-crd` `upstream.yaml` and annotations on its latest chart version

adamkpickering:
There was a bug in `partner-charts-ci` (see https://github.com/rancher/partner-charts-ci/pull/30) that resulted in the latest version of the `ezd-crd` chart not having the `catalog.cattle.io/hidden` annotation. This PR fixes that. It also adds `Hidden: true` to the `upstream.yaml` of `linux-polska/ezd-crd`.


charts/linux-polska/ezd-crd/1.3.1/Chart.yaml
```diff
@@ -1,6 +1,7 @@
 annotations:
   catalog.cattle.io/certified: partner
   catalog.cattle.io/display-name: CRDs for LP Backend
+  catalog.cattle.io/hidden: "true"
```
recena, line 4: What is the meaning of this annotation?

adamkpickering: It causes the chart to not be shown by the UI. In the partner charts repository it is used in only the `linux-polska/ezd-crd` package (this one). I think it is used here because it is effectively a CRD chart for `linux-polska/ezd-backend`.

recena (approved):

PR #1076, opened 2024-10-10
Deprecate the stackstate-k8s-agent in favor of the SUSE Observability agent

craffit:
StackState was rebranded to SUSE observability and the agent chart to SUSE Observability Agent

The SUSE Observability Agent chart is distributed as part of the rancher prime chart repository, so this chart is now deprecated.

### DIFF
--- a/packages/stackstate/stackstate-k8s-agent/upstream.yaml
+++ b/packages/stackstate/stackstate-k8s-agent/upstream.yaml
@@ -1,6 +1,7 @@
 HelmRepo: https://helm.stackstate.io
 HelmChart: stackstate-k8s-agent
 Vendor: StackState
+Deprecated: true
 DisplayName: StackState Agent
 ChartMetadata:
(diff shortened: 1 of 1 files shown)


recena (approved):

adamkpickering:
Hi @craffit and @recena, this isn't the right way to deprecate a chart. The right way is to use the `partner-charts-ci deprecate` command. This sets `Deprecated: true` in `upstream.yaml`, but also adds `deprecated: true` to the `Chart.yaml` of each chart version for that chart. This second part is important because it makes the chart show up as deprecated in the rancher UI.

@craffit is there any chance you could submit another PR for this that uses `partner-charts-ci deprecate`?

I will update the documentation for the `Deprecated` field (and the `Hidden` field, where this also applies) so that this doesn't happen in the future.

adamkpickering:
Certainly. @craffit please disregard my previous request. I'll have another PR up shortly

PR #1079, opened 2024-10-23
Rename `linkerd` vendor to `buoyant`

adamkpickering:
Linkerd is maintained by Buoyant, which is a SUSE partner. That means that the linkerd-related charts should be provided by the `buoyant` vendor, not the `linkerd` vendor. This PR makes this change. It is the result of running the following commands:
```shell
git mv assets/linkerd assets/buoyant
git mv charts/linkerd charts/buoyant
git mv packages/linkerd packages/buoyant
sed -i 's/assets\/linkerd/assets\/buoyant/g' index.yaml
```
I noticed that the `Files changed` view is not very helpful, but I'm not sure how to make the PR smaller. For what it's worth, I did test the change in a local Rancher cluster and everything worked as expected.


recena:
@adamkpickering There are conflicts.

recena (approved):

PR #1090, opened 2024-11-26
Deprecate `pixie/pixie-operator-chart`

adamkpickering:
This PR deprecates `pixie/pixie-operator-chart` in preparation for its removal according to the [deprecation and removal policy](https://github.com/rancher/partner-charts?tab=readme-ov-file#deprecation-and-removal-policy). It is being deprecated because it is now a CNCF project, so would be better provided via the [Rancher Prime Application Collection](https://apps.rancher.io/).


troytop (approved):
Looks good.

PR #1094, opened 2025-01-02
Add portworx

pragrawal-px:


assets/icons/portworx.png
troytop, line 1: Consider replacing this with Portworx_Logo_Vertical_Color.png as it will look better on the tile in the Charts view.

troytop (changes requested):
Would look better with vertical / square fit logo.

Screenshot 2025-01-02 at 11 49 03 AM

adamkpickering (commented):
Looks good to me, but I agree with @troytop - a different icon would look better.

pragrawal-px:
@troytop @adamkpickering 
If you notice the upstream.yaml, I have overridden the default icon with a new vertical Icon. But for some reason it is not reflecting.
With the last commit, I update the payload with our new Chart version 5.1.1. The changes I made on questions.yaml were properly updated. But the Icon is not changing.

Can you suggest how to fix it. Is there any other way to set the Icon field that is show on the catalog tile

adamkpickering:
Hi @pragrawal-px, `partner-charts-ci` only downloads the icon if the icon for that package is not already present in `assets/icons`. So if you remove `assets/icons/portworx.png`, it will be replaced the next time `partner-charts-ci update` downloads and integrates a new version of `portworx/portworx`.

But presumably there are not yet any newer versions of `portworx/porworx` available from upstream. So in order to replace the icon, you have to remove the artifacts associated with this package from your branch. This will cause `partner-charts-ci` to re-download these artifacts, and the icon along with them. The easiest way to do this is:

1. Back up the contents of `packages/portworx/portworx` somewhere on your system.
2. Do a hard reset of your branch to `main-source` (i.e. check out `portworx-add` and then run `git reset --hard main-source`).
3. Recreate `packages/portworx/portworx`.
4. Run `bin/partner-charts-ci update --commit` again.

This should replace the icon with whatever you've set it to in `upstream.yaml`. Note that this will leave only the latest version of portworx in your PR. I hope that helps!

pragrawal-px:
Thank you @adamkpickering for the explanation and the steps. I have performed the same and made the changes.
Here is how it looks now. Please let me know if any other is change is required.

image

adamkpickering (approved):
Looks good to me. Thank you for the contribution :smile:

image

PR #1095, opened 2025-01-08
Remove citrix packages

adamkpickering:
In #1063, #1064, #1065 and #1066 the packages from the `citrix` vendor were deprecated. According to the [deprecation and removal policy](https://github.com/rancher/partner-charts?tab=readme-ov-file#when-and-how-are-helm-charts-removed-from-this-repository) a package must be deprecated for a minimum of 3 months before it is removed. This time period has elapsed for these packages. Accordingly this PR removes them.

Please see the PRs linked above as to why these charts are being removed, and what to do if you are using them.


troytop (approved):
Looks good. Thanks!

PR #1106, opened 2025-02-18
Add Gitea package

techknowlogick:


troytop (approved):
Tested from the commitgo fork, and was able to see the tile and deploy Gitea with some basic options using the form view.

packages/commitgo/gitea/upstream.yaml
```diff
@@ -0,0 +1,7 @@
+HelmRepo: https://dl.gitea.com/charts
+HelmChart: gitea
+Vendor: CommitGo, Inc.
+DisplayName: Gitea Source Control and CI/CD Server
+ChartMetadata:
+  kubeVersion: '>=1.24-0'
+  appVersion: 1.23.3
```
adamkpickering, line 7: It is best not to include `appVersion` here. The reason: new versions of the helm chart will automatically be downloaded from the source you've specified, and integrated into the partner charts repo. When this happens, this `appVersion` will override `appVersion` from the upstream chart, which will cause it to be inaccurate unless you update this `appVersion` before every upstream chart release.

The place to set `appVersion` is in the upstream chart. Does the upstream chart not set this?

adamkpickering (changes requested):

charts/commitgo/gitea/10.6.0/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @justusbunsi @pat-s
```
recena, line 1: Ditto

techknowlogick:
Thanks @recena  I've updated to a version that has an appropriate helmignore file :)

adamkpickering (approved):

adamkpickering:
image

PR #1110, opened 2025-03-24
Added Joget package

jogetameerizwan:
Added Joget package


packages/joget/joget/upstream.yaml
```diff
@@ -0,0 +1,6 @@
+HelmRepo: https://jogetworkflow.github.io/helm-joget/charts
+HelmChart: joget
+Vendor: joget
+DisplayName: Joget DX
+ChartMetadata:
+  kubeVersion:  '>=1.21-0'
```
adamkpickering, line 6: It is best not to set `kubeVersion` in `upstream.yaml`. If the upstream chart sets `kubeVersion` to something other than what you have here, this will still override it, even if it is now wrong. Yes, it would be possible to remember to change this override when that happens, but in practice it will get forgotten.

If the upstream value is wrong it should be fixed in the upstream chart. Is this not possible?

packages/joget/joget/upstream.yaml
```diff
@@ -0,0 +1,6 @@
+HelmRepo: https://jogetworkflow.github.io/helm-joget/charts
```
adamkpickering, line 1: While we're at it, it is best practice to add a newline to the end of these files (github is highlighting it as an issue for me).

adamkpickering (changes requested):
Looks good, just a couple of minor comments.

packages/joget/joget/upstream.yaml
jogetameerizwan: Yes, it is possible. Will remove it `kubeVersion` from `upstream.yaml`

jogetameerizwan: Removed the `kubeVersion`

jogetameerizwan:
Updated the PR: Removed `kubeVersion` and added newline.

packages/joget/joget/upstream.yaml
```diff
@@ -0,0 +1,4 @@
...
+HelmChart: joget
+Vendor: joget
```
adamkpickering, line 3: Do you want to display `Joget` (i.e. with a capital J) as the vendor name in the Rancher UI? If so, you should have `Vendor: Joget`. Otherwise there is no point in setting `Vendor: joget`, since `joget` can be gotten from the package path.

index.yaml
```diff
@@ -20961,6 +20961,31 @@ entries:
     urls:
     - assets/intel/intel-device-plugins-sgx-0.26.1.tgz
     version: 0.26.1
+  joget:
+  - annotations:
+      catalog.cattle.io/certified: partner
+      catalog.cattle.io/display-name: Joget DX
+      catalog.cattle.io/kube-version: '>=1.21-0'
```
adamkpickering, line 20968: Looks like you didn't re-generate the helm chart after making changes to `upstream.yaml`. The easiest way to do this is to start fresh with the `upstream.yaml` you want, and then go through the `partner-charts-ci update ...` again. That way you know that the included chart matches what is in `upstream.yaml`.

adamkpickering (changes requested):

index.yaml
jogetameerizwan: had to actually run `remove` the charts/package before running the `update` again

jogetameerizwan:
Updated vendor in `upstream.yaml`  and refreshed `index.yaml`

adamkpickering (approved):

PR #1113, opened 2025-04-25
Update `f5/nginx-ingress` icon, and update `README.md` with info on how to update a chart's icon

adamkpickering:
I tested the icon in a development version of Rancher and it looked good. I'd post a screenshot but it seems my screenshot utility is broken.


troytop (approved):
Thank you! Looks good to me.

README.md
```diff
@@ -70,6 +70,14 @@ is useful for this; for more information please see the
 [`partner-charts-ci` documentation](https://github.com/rancher/partner-charts-ci?tab=readme-ov-file#4-run-partner-charts-ci-update).
```
recena, line 72: This empty line is not needed.

adamkpickering: The readme currently has two empty lines after each section. I like doing it this way because it makes the file easier to read in a text editor. I can change this if you want, but it doesn't have any effect on the HTML version of the file.

PR #1123, opened 2025-06-23
Nirmata n4k rancher release 1.13 validation

mkk0083:
This PR adds the  N4K (Nirmata for Kyverno) Helm chart under packages/nirmata/n4k.
Includes:
App metadata and branding
app-readme.md
Chart.yaml and values.yaml
Validation tested with local Rancher instance (rke version v1.8.4)

Let me know if any additional steps are needed before merge.
Thanks,

Mayur Kotamraju


adamkpickering (changes requested):
One moment, need to check if Nirmata is a SUSE partner.

PR #1144, opened 2026-04-01
chore(dh2i): deprecate dxoperator v1 and promote dxoperator-v2 to featured

nicholasSUSE:
## Summary
- Remove `catalog.cattle.io/featured` from dh2i/dxoperator (v1 legacy)
- Add `catalog.cattle.io/featured: "1"` to dh2i/dxoperator-v2
- Set `Deprecated: true` in dxoperator upstream.yaml to stop nightly bot from fetching new versions
- Run `partner-charts-ci deprecate` on dxoperator v1 to mark all chart versions (1.0.1–1.0.5) as deprecated
- Restore dxoperator-v2 1.0.0 assets overwritten by the deprecate tool, which unconditionally fetches from upstream (DH2i replaced their dxoperator upstream with v2 content)


troytop (approved):
Looks good to me.

Issue #1152, opened 2026-05-19
bug: kamaji chart v1.0.0 - missing etcd-server ClusterIP service breaks etcd backup


CYBERBUGJR:
## Summary

The `clastix/kamaji` chart v1.0.0 distributed through the Rancher partner catalog has a bug in the bundled etcd deployment: the `etcd-server` ClusterIP service is never created, even though the auto-generated TLS certificates include `etcd-server.{namespace}.svc.cluster.local` in their SANs.

This makes etcd backup (and any `etcdctl` access via a stable service endpoint) impossible with TLS verification.

## Upstream issue

Reported to Clastix: https://github.com/clastix/kamaji/issues/1157

## User impact

Users who deploy Kamaji via the Rancher UI partner catalog and attempt to back up etcd using standard procedures will see:

```
x509: certificate is valid for etcd-0.etcd.kamaji-system.svc.cluster.local, ...,
etcd-server.kamaji-system.svc.cluster.local, not etcd.kamaji-system.svc.cluster.local
```

## Suggested action

Once Clastix releases a patched chart version (or a new version that properly deprecates the bundled etcd in favor of `kamaji-etcd`), please update the partner catalog entry.

In the meantime, consider adding a note in the chart's `app-readme.md` warning that the bundled etcd backup is not functional and pointing users to the `kamaji-etcd` standalone chart.

CYBERBUGJR:
## Update: upstream response + recommended action

The upstream Clastix team has responded on the upstream issue
(https://github.com/clastix/kamaji/issues/1157#issuecomment-4514981921):

> That Helm Chart is deprecated and we don't ship anymore fixes on that since our
> monetization model for the Open Source is granting access to LTS stable artefacts,
> such as OCI and OCI-based Helm Charts. Furthermore, we decoupled Kamaji from
> Kamaji etcd as a sub chart.

**The `clastix/kamaji` chart v1.0.0 will not receive a patch for this bug.**

### Verification on `kamaji-etcd`

I checked the standalone `kamaji-etcd` chart (https://github.com/clastix/kamaji-etcd):
- It correctly creates an `etcd-client` ClusterIP service (`etcd_client_service.yaml`)
- The auto-generated cert SANs include `{fullname}-client.{namespace}.svc.cluster.local`,
  matching the service name
- etcdctl and backup access via the stable ClusterIP endpoint works correctly

The bug does **not** affect the standalone `kamaji-etcd` chart.

### Requested action from Rancher

[...]

PR #1159, opened 2026-07-10
removing kamaji

nicholasSUSE:
Removing Kamaji


briandowns (approved):

troytop (approved):
Removing because the chart is broken (see https://github.com/rancher/partner-charts/issues/1152). The upstream kamaji chart was already deprecated, so we're shortcutting the normal deprecation timeline.

